PR #4, opened 2016-11-27
(fix) remove poller from webpack-dev-server watch options

### LINKED ISSUE #3: Node process uses a lot of CPU when in dev mode
Nice project, very inspiring as I'm building a set of custom components too :)

On thing I noticed when running it using `npm start` is that there was a node process that never got below 70% CPU usage.

I found that [this line](https://github.com/swimlane/swui/blob/master/config/webpack.dev.js#L17) causes that behavior. The app seemed to keep working when I uncommented it, so I'm not sure if it's needed.

beeman:
Fixes #3


amcdnl:
Thanks so much!

PR #25, opened 2017-03-21
feature(icons): new icons

Hypercubed:


Hypercubed:
image
image
image
image
image

Hypercubed:
image

amcdnl:
I believe there is existing icons for many of these. We should update the existing ones to the new ones rather than duplicate.

PR #32, opened 2017-04-05
New feature, icon registry

Hypercubed:
With this new feature icons aliases, including aliases for stacked icons, can be registered using the `iconRegisteryService` then included in templates using `ngx-ui`.

Example form demo:

image


amcdnl:
🎉

PR #44, opened 2017-05-01
style: update current styles

Hypercubed:


amcdnl:
Don't manually add the CSS HEX equivalent manually ... we auto generate these in other places at build time, just include these in that.

PR #46, opened 2017-05-05
Update styles

Hypercubed:


src/styles/typography/font.scss
```diff
@@ -91,6 +91,6 @@ span, p, a {
 }
 
 a {
-  color: $color-blue;
+  color: $color-blue-med;
```
amcdnl: Where does `$color-blue-med` come from? I do not know this reference, i'm suspecting you mean `color-blue-500`?

Hypercubed: It was there until you deleted it: https://github.com/swimlane/ngx-ui/commit/8711827c3aea84a88500ab2ff515b1e39a082df5#diff-052415be6a417fa205cfe86eaf6410d1L15   😄

PR #60, opened 2017-06-05
allow disableDropdown in ngx-select

Hypercubed:


amcdnl:
What is the driver behind this?

Hypercubed:
I wanted a way to disable the drop down while maintaining ability to see and remove tags...  an example could be to enable and disable selected values (length, values, etc).  Mock example:

```
<ngx-select [disableDropdown]="items.length === 1 && item.id === 'all'">
  ...
</ngx-select>
```

amcdnl:
LGTM

PR #61, opened 2017-06-08
Linting

Hypercubed:


src/components/drawer/drawer.service.ts
```diff
@@ -47,11 +47,9 @@ export class DrawerService extends InjectionRegisteryService {
     bindings = super.assignDefaults(bindings);
 
     if(!bindings.inputs.zIndex) {
-      if (this.overlayService.instance) {
```
amcdnl: I don't agree with this rule, I thought I had turned it off in latest version of lint rules.

PR #62, opened 2017-06-08
run hotkey callbacks in zone

Hypercubed:


src/components/hotkeys/hotkeys.service.ts
```diff
@@ -7,6 +7,8 @@ const hotkeys = {};
 const hotkeyChangedSource = new Subject();
 const isMac = window.navigator && window.navigator.platform.indexOf('Mac') !== -1;
 
+const ngZone = new NgZone(false);
```
amcdnl: Don't create a new zone, inject in the angular one.

Hypercubed: In fact, that line is not needed at all and will not work.  The zone for teh call back is set in two places:

1) For the service it is injected: https://github.com/Hypercubed/ngx-ui/blob/d73c0550f0eab83257b0fde5e50fa9415ffdcd90/src/components/hotkeys/hotkeys.service.ts#L146
2) For the `Hotkey` decorator, we cannot inject, so we create a new zone here: https://github.com/Hypercubed/ngx-ui/blob/d73c0550f0eab83257b0fde5e50fa9415ffdcd90/src/components/hotkeys/hotkeys.service.ts#L131

src/components/hotkeys/hotkeys.service.ts
```diff
@@ -143,14 +141,14 @@ export function Hotkey(key, description: string, options?: any) {
...
 export class HotkeysService {
-  constructor(private ngZone: NgZone) {}
-
   hotkeys = hotkeys;
   suspend = _suspend;
   activate = _activate;
   deregister = _deregister;
   changeEvent = hotkeyChangedSource.asObservable();
 
+  constructor(private ngZone: NgZone) {}
```
amcdnl: This only works now if you use the service, so decorators won't work.

Hypercubed: Decorators work because of this: https://github.com/Hypercubed/ngx-ui/blob/d73c0550f0eab83257b0fde5e50fa9415ffdcd90/src/components/hotkeys/hotkeys.service.ts#L131

You can test this in the demo by using the theme switching hotkeys.

PR #63, opened 2017-06-08
Updated and add various icons

Hypercubed:


src/assets/svgs/remove-users.svg
```diff
@@ -0,0 +1,12 @@
+<svg width="19" height="12" viewBox="0 0 19 12" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+<title>remove-users</title>
+<desc>Created using Figma</desc>
```
amcdnl: Lets make sure and remove metadata in the future.

PR #71, opened 2017-06-23
Consistant alignment, colors, and font sizes for ngx-select and ngx-inputs

Hypercubed:


src/components/input/input.component.scss
```diff
@@ -23,7 +23,7 @@ $input-icon-color: $color-grey-300;
 
   .ngx-input-wrap {
     position: relative;
-    display: inline-block;
+    display: block;
```
amcdnl: inputs needs to be able to sit beside each other.

demo/app/app.template.html
```diff
@@ -372,7 +558,7 @@ <h3 class="style-header">Align Bottom</h3>
             </ngx-select>
           </td>
         </tr>
-      </table>
+      </table -->
```
amcdnl: Why'd you comment this out?

Hypercubed: Should be removed, only baseline alignment is relevant.

src/components/input/input.component.scss
Hypercubed: This is the `.ngx-input-wrap` class.  This should not effect the ngx-input elements, which were and still are, block elements.

amcdnl:
Rebase and go ahead and merge @Hypercubed

PR #78, opened 2017-06-30
Updated preloader to respond during JS load, fixes #68

### LINKED ISSUE #68: Spinner animation is skipping when heavy processing is done
The animation of the spinners is choppy and skips when there is heavy processing done at the same time. This can be noticed when you have the spinner running while the page is still loading.

@Hypercubed found out that this is caused by the problems described in this article:
https://blog.keanulee.com/2014/10/20/the-tale-of-three-spinners.html

We need to rewrite the spinner animations not to use dasharrays, and use something more performant instead.

Hypercubed:
https://www.screencast.com/t/xw2Bt9SP

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-_(none)_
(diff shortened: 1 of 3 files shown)


amcdnl:
LGTM

PR #82, opened 2017-07-07
Fix/numeric inputs

Hypercubed:


src/components/input/input.component.ts
```diff
@@ -199,7 +199,13 @@ export class InputComponent implements OnInit, AfterViewInit, ControlValueAccess
   }
 
   get focusedOrDirty(): any {
-    return this.focused || (this.value && this.value.length);
+    if (this.focused) {
+      return true;
+    }
+    if (this.value === 'string') {
```
amcdnl: don't you mean typeof?

PR #83, opened 2017-07-07
Feature: add `pauseOthers` and `unpauseOthers` to hotkeys service

Hypercubed:


demo/app/app.component.ts
```diff
@@ -525,6 +525,15 @@ function moo() {
     this.dialogMngr.create(options);
   }
 
+  openDialogAndPauseHotkeys(options) {
+    const dlg = this.dialogMngr.create(options);
+    this.hotkeysService.pauseOthers(dlg.instance);
+    dlg.instance.close.subscribe(() => {
+      console.log('close');
```
amcdnl: Remove console here

amcdnl:
Whats the use case behind this?

Hypercubed:
When opening a dialog we should disable hotkeys, these keys should be reenabled after the dialog is closed.  This implementation allows other hotkeys to be disabled in bulk, possible excluding the current component.  This implementation also allows for nesting of pause and unpause method calls.

PR #100, opened 2017-07-21
Fix: mod keys maps to ctrl when not mac

Hypercubed:


src/components/hotkeys/hotkeys.service.ts
```diff
@@ -25,8 +25,8 @@ function _getDisplay(combo) {
   const result = [];
 
   for(const k of keys) {
-    if(k === 'mod' && isMac) {
-      result.push(map.command);
+    if(k === 'mod') {
+      result.push(isMac ? map.command : 'ctrl');
```
amcdnl: When using `mod` as the key accessor, this should be handled automatically by the Mousetrap library

Hypercubed: It is...  but without this change the help text still displays "mod".

amcdnl: Ah, this is just for display. Got it.

PR #108, opened 2017-08-29
Feature(ngx-tabs): add optional ngxIfTabActive directive

Hypercubed:


src/components/tabs/tab.component.ts
```diff
@@ -1,14 +1,23 @@
...
-    <div [hidden]="!active">
-      <ng-content></ng-content>
-    </div>
+    <div *ngIf="template; then template_content else ng_content"></div>
+    <ng-template #template_content>
+      <div *ngIf="active">
+        <ng-container [ngTemplateOutlet]="template.templateRef"></ng-container>
+      </div>
+    </ng-template>
+    <ng-template #ng_content>
```
amcdnl: I don't like the name `#ng_content`, say something like `container` or something like that.

Hypercubed: I'll change this and the outdated note above.  Are you happy with the `ngxIfTabActive` directive name?

amcdnl: Ya thats good

PR #109, opened 2017-09-21
Improved performance and AoT for ngx-inputs

Hypercubed:


src/components/input/input.component.ts
```diff
@@ -22,7 +22,7 @@ const INPUT_VALUE_ACCESSOR = {
   template: `
     <div
       class="ngx-input-wrap"
-      [ngClass]="getCssClasses">
+      [ngClass]="{ 'ng-invalid': this.isInvalid, 'ng-touched': this.isTouched, 'ng-valid': isValid }">
```
amcdnl: I'd avoid inline `ngClass` objects and whatnot. Not performant.

marjan-georgiev: Maybe use the `[class.ng-invalid]="this.isInvalid"` syntax instead?

Hypercubed: Done... thanks.

PR #117, opened 2017-09-29
(feat): Restyle dialog

marjan-georgiev:
- Moves the close button outside of the dialog
- Adds `class` input to dialog component
- Adds 725, 775, 825, and 875 values to grey and blue-grey color pallettes
- Adds styles for dialog with class `wizzard`
- Adds wizzard demo to dialog demo page


demo/app/app.template.html
```diff
@@ -2741,6 +2741,121 @@ <h3 class="style-header">Dialog</h3>
...
+            </ngx-tab>
+          </ngx-tabs>
+
+          <div class="ngx-dialog-footer">
+            <div class="wizzard-nav">
+              <button type="button" class="btn btn-bordered">
+                <i class="icon icon-arrow-left has-text"></i>Prev
+              </button>
+              <button type="button" class="btn btn-bordered">
+                Next <i class="icon icon-arrow-right"></i>
```
Hypercubed: May want to make these buttons functional.

PR #119, opened 2017-09-29
(feat): Add showCaret input to dropdowns

marjan-georgiev:
Adds `showCaret` input to dropdown component

### DIFF
--- a/demo/app/app.template.html
+++ b/demo/app/app.template.html
@@ -3773,7 +3773,7 @@
         </ngx-section>
 
         <ngx-section class="shadow" [sectionTitle]="'Dropdown Content'">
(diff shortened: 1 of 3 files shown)


demo/app/app.template.html
```diff
@@ -3773,7 +3773,7 @@ <h3 class="style-header">Dropdown</h3>
         </ngx-section>
 
         <ngx-section class="shadow" [sectionTitle]="'Dropdown Content'">
-          <ngx-dropdown>
+          <ngx-dropdown [showCaret]="true">
```
Hypercubed: Need to add this to the demo code below.

PR #120, opened 2017-09-29
ngx-nag

Hypercubed:


amcdnl:
@Hypercubed - Can you provide a description of what this is and the use case?

Hypercubed:
This is a persistent "drawer" and notification area similar in design to Angular Material's SnackBar.  It provides a notification of state with the ability to expand for more details.  It could be used for displaying events that must be reacted to or perhaps as a shopping cart.

Please see the following video:

https://www.screencast.com/t/RytdlTlgic

src/styles/fonts/icons.scss
```diff
@@ -1,8 +1,14 @@
 @font-face {
   font-family: "icon";
+<<<<<<< HEAD
```
marjan-georgiev: Unresolved merge conflict. This breaks the build locally.

PR #122, opened 2017-10-03
(fix): Fix ExpressionChangedAfterItHasBeenCheckedError error

marjan-georgiev:


src/components/input/input.component.ts
```diff
@@ -235,6 +233,15 @@ export class InputComponent implements OnInit, AfterViewInit, ControlValueAccess
       this.inputModel.touched;
   }
 
+  get getCssClasses(): any {
```
amcdnl: This is gonna run on each CD cycle. The previous was more performant, notice there is a `this` on the previous commit, wonder if that could be doing it.

marjan-georgiev: The previous also ran on each CD cycle (checked with console log). All getters do. Tried without the `this` too, nothing fixed the error.

PR #124, opened 2017-10-04
SPT-1267: Fixed calendar and time input styles

Hypercubed:


src/components/calendar/calendar.component.scss
```diff
@@ -1,33 +1,43 @@
 @import "colors/index";
 
-$calendar-bg: $color-blue-grey-600;
-$calendar-title-bg: $color-blue-grey-700;
+$calendar-bg: $color-blue-grey-800;
+$calender-border: #2D3544;
```
marjan-georgiev: Can we replace these with existing colors? If the exact ones are not present, let's use the closest ones.

src/components/calendar/calendar.component.ts
```diff
@@ -1,7 +1,9 @@
 import { Component, Input, Output, EventEmitter, forwardRef, OnInit, ViewEncapsulation } from '@angular/core';
 import { ControlValueAccessor, NG_VALUE_ACCESSOR } from '@angular/forms';
 import * as moment from 'moment';
-import { getMonth } from './calendar-utils';
+import { getMonth, CalenderDay, Month } from './calendar-utils';
+
+type Classes = string[]|Set<string>|{[klass: string]: any};
```
marjan-georgiev: is this used in this file?

Hypercubed: No longer used, should be removed.

PR #127, opened 2017-10-10
Chore: allow submodule-imports in tslint rules

Hypercubed:


amcdnl:
You should add this to the global tslint rules

PR #129, opened 2017-10-12
Fix(dropdown): Use HostBinding and toggle event instead of renderer

Hypercubed:


amcdnl:
This is done for performance. The body listener only listens when it's open.

Hypercubed:
@amcdnl I understand, but it does not work in some instances.

amcdnl:
I strongly urge against a listener on the body that is always on.

Hypercubed:
The existing implementation, in some cases, fails to trigger change detection after the call back.  An alternative is to add `this.cd.markForCheck` to each callback.

amcdnl:
I think that would be much better solution.

src/components/dropdown/dropdown.component.ts
```diff
@@ -46,33 +46,42 @@ export class DropdownComponent implements AfterContentInit, OnDestroy {
...
       }
     }
   }
 
   onToggleClick(ev): void {
     if(!this.dropdownToggle.disabled) {
       this.open = !this.open;
+
+      if (this.open) {
+        this.documentListener = this.renderer.listen(
```
amcdnl: You should look at using RX for listening to the stream of click events.

Hypercubed: I'll look into it... should be done across components.

PR #130, opened 2017-10-18
Update tooltip styles

marjan-georgiev:


amcdnl:
You should put screenshots when u update styles.

marjan-georgiev:
image

image

src/components/tooltip/tooltip.component.scss
```diff
@@ -27,7 +26,7 @@ $popover-spacing: 10px;
     border: 1px solid $popover-border;
     box-shadow: $shadow-3;
     font-size: 13px;
-    padding: 4px;
+    padding: 10px;
```
Hypercubed: I think these should be in em... in this case probably 0.667em.

PR #143, opened 2017-12-12
Update chg to get rid of unsecure version of semver

marjan-georgiev:
chg depends on semver ~2.2.1, which is tagged as high severity security vulnerability by github.


Hypercubed:
Let's close this.... the fix is to bump chg to 0.3.3.

PR #146, opened 2017-12-26
New ngx-ui icons

Hypercubed:
This change:

* Adds updated SVG and font icons generated via nucleo
* New icons are namespaced with nxg.
* Leaves legacy icons without namespace, still used by some components.
* Removes icon build steps
* Updates demo to use new icons.


src/assets/icons/iconfont/less/nucleo.less
```diff
@@ -0,0 +1,166 @@
+/* --------------------------------
```
amcdnl: Using LESS and SASS? :(

PR #150, opened 2018-01-19
add new icons, remove fill spec from a few SVGs, rename 'plan' to 'plane' and 'info-fulled' to 'info-filled'

mcwenger:
It looks like Nucleo has removed the fill from a number of SVGs – I think this is what we want, so I committed the change.


marjan-georgiev:
👍

PR #157, opened 2018-02-08
Add full screen dialog styles

marjan-georgiev:


src/components/dialog/dialog.component.scss
```diff
@@ -107,6 +107,32 @@ $dialog-body-color: $color-grey-300;
     }
   }
 
+  &.ngx-dialog-full-screen {
+    width: 100vw;
+    height: 100%;
```
amcdnl: You probably want 100vh and to scroll that inside.

PR #158, opened 2018-02-13
Added radio button and radio button group

Hypercubed:


amcdnl:
This could be cleaner, checkout how angular does it - https://github.com/angular/angular/blob/master/packages/forms/src/directives/radio_control_value_accessor.ts#L24

Hypercubed:
@amcdnl I followed a combination of ngx-ui patterns and angular material.

nhhockeyplayer:
is there a way to set a radio button 'checked' onInit ?
I tried to set the input and it seems like a dead end

any tips are appreciated

none of the radios are coming up with an initial one checked for initialized default radio button state for any radio group

thanks

nhhockeyplayer:
this is terrible, cant even set a radio checked
can someone offer feedback
looks like the details are all modeled in scss and something isnt holding water

PR #162, opened 2018-02-26
Add quotes around URLs: source-sans.scss

bostondevin:
url strings need to be quoted to work with latest angular CLI


marjan-georgiev:
Fixed as part of https://github.com/swimlane/ngx-ui/pull/206. Thanks!

PR #163, opened 2018-02-26
quote url strings: lato.scss

bostondevin:
url stings need to be quoted to work with latest angular CLI


marjan-georgiev:
Closing as Lato is no longer used.

PR #164, opened 2018-02-26
quote url strings

bostondevin:
Quote URL strings to work with latest angular CLI


Hypercubed:
This file is auto-generated.

PR #177, opened 2018-05-17
Add getRootViewContainerNode to InjectionService

Hypercubed:


src/services/injection.service.ts
```diff
@@ -19,7 +19,8 @@ import {
  */
 @Injectable()
 export class InjectionService {
-  private _container: ComponentRef<any>;
+  static _container: ViewContainerRef;
```
marjan-georgiev: let's rename this to something else (globalContaier, staticContainer) please.

Hypercubed: Will do....

PR #204, opened 2018-10-26
Angular 7 Upgrade and Angular CLI Migration

metalex9:
Upgrades library to Angular 7 and migrates entire project to use the Angular CLI.

Because ngx-ui uses global styles, you'll need to do `npm run build-pkg` to build ngx-ui (this will run `ng build @swimlane/ngx-ui` as well as preparing the styles).

There were a few changes I had to make to components to make ng-packagr happy.


metalex9:
Looks like I need to fix the Travis CI config

metalex9:
Travis CI is passing again 🎉

PR #211, opened 2018-12-26
added [caseSensitive] input to ngx-select

Hypercubed:


marjan-georgiev:
Looks good. Can we rename the input to 'filterCaseSensitive'?

PR #213, opened 2019-01-10
Support timezone display for ngx-date-time

Hypercubed:
Maybe a breaking change.... adds moment-timezone as a peer.


marjan-georgiev:
@Hypercubed the build fails. Could you take a look please?

Issue #222, opened 2019-01-29
Refactor demo - split into pages


marjan-georgiev:
The demo page is becoming too big. Split each demo section into a separate page and add @angular/router.

PR #224, opened 2019-01-31
Refactor demo

haeminn:
#222 
- Broke down each page into component
- Added router


src/app/app.module.ts
```diff
@@ -4,16 +4,144 @@ import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
 import { FormsModule } from '@angular/forms';
 import { NgxDatatableModule } from '@swimlane/ngx-datatable';
 
-import { AppComponent } from './app.component';
-import { CounterComponent } from './counter.component';
 import { MomentModule } from 'ngx-moment';
-import { NgxUIModule } from '../../projects/swimlane/ngx-ui/src/public_api';
+import { NgxUIModule, DateTimeComponent } from '../../projects/swimlane/ngx-ui/src/public_api';
```
marjan-georgiev: DateTimeComponent does not seem to be used here, we can remove it.

src/app/app.component.scss
```diff
@@ -31,8 +31,8 @@
   }
 
   .main {
-    display: flex;
-    flex-grow: 1;
+    // display: flex;
```
marjan-georgiev: If these are no longer needed, let's delete them.

src/app/components/hotkeys-page/hotkeys-page.component.html
```diff
@@ -0,0 +1,124 @@
+<div class="page">
```
marjan-georgiev: Since we're wrapping every template with this page div, it can be moved into app.template.html and wrap the router-outlet element with it, then remove the page class from components that have it on their root section.

src/app/app.module.ts
```diff
@@ -4,16 +4,144 @@ import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
...
 
 import { HttpClientModule } from '@angular/common/http';
 import { FileUploadModule } from 'ng2-file-upload';
 import { FlexLayoutModule } from '@angular/flex-layout';
+
+import { RouterModule, Routes } from '@angular/router';
+
+// Components
+import { AppComponent } from './app.component';
+import { CounterComponent } from './counter.component';
```
marjan-georgiev: The counter component is a helper component for the Tabs page, so let's move it into `/components/tabs-page`

src/app/app.component.ts
```diff
@@ -19,6 +19,7 @@ import {
 } from '../../projects/swimlane/ngx-ui/src/public_api';
```
marjan-georgiev: This file has some unused imports that we can remove now that most of it's contents have been moved out.

package.json
```diff
@@ -31,6 +31,7 @@
     "@angular/platform-browser": "~7.0.0",
     "@angular/router": "~7.0.0",
     "@swimlane/ngx-datatable": "^14.0.0",
+    "@swimlane/ngx-ui": "^25.1.0",
```
marjan-georgiev: Is this needed?

src/app/components/hotkeys-page/hotkeys-page.component.html
haeminn: I noticed that some components use `<section>` not `<div>`

src/app/app.module.ts
haeminn: Don't you have to register `CounterComponent` in a module though?

marjan-georgiev: Yes, just commenting here for the file location.

haeminn: Okay I understand what you mean now

package.json
haeminn: I followed the installation instruction on its README:
`npm i @swimlane/ngx-ui --S` 

I can remove this though

src/app/components/hotkeys-page/hotkeys-page.component.html
marjan-georgiev: Yes, should be the same.

marjan-georgiev:
👍

PR #236, opened 2019-02-15
Fixes bug where changing the direction of ngxSplit will break the layout

marjan-georgiev:
- Refactors the split component into a directive, and fixes a bug where changing the direction of ngxSplit will break the layout
- Adds direction change demo


marjan-georgiev:
@Hypercubed pushed an update.

Issue #233, opened 2019-02-14
Add gradient button class


marjan-georgiev:
Add a `btn-primary-gradient` class to the list of button classes that looks like the button on the right here:

![image](https://user-images.githubusercontent.com/8687/52816830-6b078180-30a2-11e9-82ea-eacffdd5ea9c.png)

Update the buttons page in the demo to showcase it

haeminn:
@marjan-georgiev 
"button on the right" as in any button class that has background color?

PR #237, opened 2019-02-15
add btn-primary-gradient class #233

haeminn:
Addresses #233.


src/app/forms/buttons-page/buttons-page.component.html
```diff
@@ -1,14 +1,14 @@
 <h3 class="style-header">Buttons</h3>
 <ngx-section class="shadow" sectionTitle="Buttons">
   <button type="button" class="btn">Default</button>
-  <button type="button" class="btn btn-primary">Primary</button>
+  <button type="button" class="btn btn-primary btn-primary-gradient">Primary</button>
```
marjan-georgiev: Let's keep the `primary` examples in there, and just add the `primary-gradient` ones in addition to those.

marjan-georgiev:
👍

PR #244, opened 2019-03-04
JSON editor support for multiple types

marjan-georgiev:
- Adds support for schemas with multiple types: 

```
jsonEditorSchema = {
    type: 'object',
    properties: {
      productName: {
        description: 'Name of the product',
        type: ['string', 'string=code', 'string=password', 'number', 'string=date', 'object']
      }
   }
}
```

- Adds ability to switch types on properties that can have different ones:
image

- Adds a new input that allows overriding the inferType methods for each type:
```
typeOverrides: any = {
    'string=code': (value: any) => {
      if (typeof value !== 'string') {
        return false;
      }
      const regex = new RegExp(/^<<(.*)>>$/, 's');
      return regex.test(value);
    }
  };

<ngx-json-editor
  [(model)]="jsonEditorModel"
  [schema]="jsonEditorSchema"
  label="Model"
  [typeCheckOverrides]="typeOverrides"
>

```

- Splits the new property dropdown into 2 columns: schema properties and additional properties

image


projects/swimlane/ngx-ui/src/lib/components/json-editor/node-types/array-node.component.html
```diff
@@ -4,10 +4,7 @@
       <ngx-dropdown [showCaret]="true">
         <ngx-dropdown-toggle>
           <div class="type-icon" tooltipPlacement="top">
-            <ngx-icon
-              [fontIcon]="dataTypeMap[(schemas[i]?.type)] ? dataTypeMap[(schemas[i]?.type)].icon : 'integration'"
-            >
-            </ngx-icon>
+            <ngx-icon [fontIcon]="getIcon(schemas[i])"> </ngx-icon>
```
duredhel13: Can we do something so we don´t call this function all the time? Perhaps have the "fontIcon" stored in a public variable in the component, and only call getIcon() on OnChanges?

projects/swimlane/ngx-ui/src/lib/components/json-editor/json-editor.component.ts
```diff
@@ -27,17 +28,27 @@ export class JsonEditorComponent implements OnChanges {
   @Input()
   label: string;
 
+  @Input()
+  typeCheckOverrides: any;
```
duredhel13: Is this always mandatory? Looks like it could perhaps be optional?

marjan-georgiev: Yes, made it optional. Is there anything in the angular template compiler that checks whether required params are passed? I haven't encountered that anywhere yet.

duredhel13: Yes, if you omit an Input on the template that is not optional the compiler should give an error. At least the aot compiler should.

projects/swimlane/ngx-ui/src/lib/components/json-editor/node-types/array-node.component.html
marjan-georgiev: Yup, I am caching them in the schema $meta.icon property now.

Issue #235, opened 2019-02-15
Avoid using native attributes (e.g. `[title]`) as a component input.


Hypercubed:
The `ngx-tab` component has a `[title]` input (https://github.com/swimlane/ngx-ui/blob/master/projects/swimlane/ngx-ui/src/lib/components/tabs/tab.component.ts#L27) when this input is used as a string (i.e. `title="The title"`) the input also becomes the hiver title.

We should rename the title input and scrub the code to ensure this is not happening elsewhere.

PR #246, opened 2019-03-06
#235

haeminn:
#235 

Modified tab and toolbar components to not use title attribute.


Hypercubed:
Can we replace `tabTitle` with `label`

PR #251, opened 2019-03-07
JSON editor array multitypes

marjan-georgiev:
Adds support for mulitple types for array items


projects/swimlane/ngx-ui/src/lib/components/json-editor/json-editor.helper.ts
```diff
@@ -184,3 +184,20 @@ export function getIcon(schema: any): string {
 
   return 'integration';
 }
+
+/**
+ * Returns a string for the schema.$meta.currentType property in the following format:
+ * string, string=code, object, etc.
+ * @param schema
+ */
+export function getCurrentType(schema): string {
```
duredhel13: It almost feels that `schema` should have an interface, or at least a Partial one.

PR #256, opened 2019-03-11
Fixing splitAreaDirective error

duredhel13:
Extra null check to avoid error.


Hypercubed:
LGTM... can you also add a comment in projects/swimlane/ngx-ui/CHANGELOG.md?

PR #264, opened 2019-03-15
Karma

Hypercubed:


projects/swimlane/ngx-ui/src/lib/components/date-time/date-time.component.ts
```diff
@@ -26,6 +25,8 @@ const DATE_TIME_VALUE_ACCESSOR = {
   multi: true
 };
 
+type Datelike = string | Date | moment.Moment;
```
marjan-georgiev: Should these changes be part of this PR? Are they related to the tests?

Hypercubed: Part of the linting... debounceable was imported but not used.  `setDialogDate` was dup of `dateSelected`

PR #273, opened 2019-03-28
Sass Improvements

metalex9:
In projects which have multiple sass compilation targets, consumers must be careful not to import style rules multiple times from different targets. A common situation is to have `@import "colors/index.scss"` in a consuming project anywhere the color variables are needed. Unfortunately, this file imports a large amount of rules for gradients, box-shadows, etc, which end up duplicated if consumers import the file from multiple compilation targets. On top of this, usually the consumer only needs access to the color variables and not the rules anyway.

I've separated variable declarations for colors from actual rules which use those colors. Variable declarations can safely be imported anywhere as they don't result in new rules in the CSS output. I've also added a new file, "colors/variables.scss", which only contains color variable declarations (including the colors for gradients and glows) but no rules. This file should be used as a replacement for "colors/index.scss" when a consumer needs access to the color variables.

As such, I've updated references to the color files in the component stylesheets. Positive effects of this can already be observed by inspecting the new generated index.css which no longer contains duplicate declarations for the ".gradient" and ".glow" rules.

This is NOT a breaking change.


projects/swimlane/ngx-ui/src/lib/styles/colors/shadow-colors.scss
```diff
@@ -0,0 +1,123 @@
+$shadow-key-umbra-opacity: 0.2;
```
duredhel13: Not sure about the naming of this file. It has more than just colors right?

metalex9: Yeah you're right. Will rename to "shadow-variables"

Issue #207, opened 2018-11-28
Unexpected value 'undefined' imported by the module 'NgxUIModule


bbougot:
Got this error while using this module in an Angular 7 project. Used to worked perfectly with 21.7.0 version (all peerDependencies and module are correctly installed and declared).

heyepe:
Experiencing the same issue on v22.0.2

ivoglent:
Any news for this issue? I like this module but can not use :(

ivoglent:
figured out by myself.
Run ng serve with option : --aot false
Example : 
```
ng --configuration=dev serve --aot false
```
For moment export error. try add :
```
"allowSyntheticDefaultImports" : true,
```
to compilerOptions of tsconfig.json.

bbougot:
Problem is I need AOT for production release.

marjan-georgiev:
This should be fixed in 25.0.0

Anhbta:
Still having same issue in 25.6.3, don't know how to fix the build.

FBNjsilvamarques:
Having same issue too. It's not possible to build with aot. Version 25.6.2

cmckni3:
Having this issue on 25.6.3 and 25.7.0.

cmckni3:
Works great in 25.8.0

PR #274, opened 2019-04-19
Fix radio button export

cmckni3:
Appears to be missed import for #207 and #217


cmckni3:
Linking against the locally built version and building with aot works for me.

marjan-georgiev:
Ah, great catch. Thank you!

PR #287, opened 2019-05-28
ngx-select required validation fix

duredhel13:
When model was inside an array `[ngModel]="[someVar]"` the invalid class was not triggered if `someVar === undefined || null || "" `


projects/swimlane/ngx-ui/src/lib/components/select/select.component.ts
```diff
@@ -123,7 +123,7 @@ export class SelectComponent implements ControlValueAccessor, OnDestroy {
 
   @HostBinding('class.invalid')
   get invalid() {
-    if (this.required && (!this.value || this.value.length < 1)) return true;
+    if (this.required && (!this.value || this.value.length < 1 || (this.value.length === 1 && !this.value[0]))) return true;
```
marjan-georgiev: `!this.value[0]`

Would we have an issue here if the value is `0`?

duredhel13: Yup, same issue in the previous statement `!this.value`. I'll compare with `undefined`.

PR #300, opened 2019-08-14
Update scrollbars

Hypercubed:
- Update scrollbar styles
- add ngx-scroll-muted


projects/swimlane/ngx-ui/src/lib/components/input/input.component.ts
```diff
@@ -317,6 +317,9 @@ export class InputComponent implements OnInit, AfterViewInit, ControlValueAccess
 
   onChange(event): void {
     event.stopPropagation();
+    if (this.type === 'number') {
```
marjan-georgiev: Looks like this sneaked in.

PR #303, opened 2019-09-09
add ability to define custom template for tab header

luisb90:
This PR adds the ability to pass in a TemplateRef for a tab label. The previous behavior of being able to pass in a string is preserved.

Screen Shot 2019-09-09 at 3 49 30 PM


projects/swimlane/ngx-ui/src/lib/components/tabs/tabs.component.html
```diff
@@ -0,0 +1,17 @@
+<section>
+  <ul class="ngx-tabs-list list-reset" [class.tabs-vertical]="vertical" [class.tabs-horizontal]="!vertical">
+    <li *ngFor="let tab of tabs" class="ngx-tab" [class.disabled]="tab.disabled" [class.active]="tab.active">
+      <button (click)="tabClicked(tab)" [disabled]="tab.disabled">
+        <ng-container
+          *ngTemplateOutlet="labelIsString(tab.label) ? labelIsStringTmpl : tab.label; context: { $implicit: tab }"
```
marjan-georgiev: Can we do the type check inline? It should improve performance.

luisb90: updated it so the right template to use is assigned on init once

PR #304, opened 2019-09-27
Add support for Ivy

marjan-georgiev:
- Updates dependencies to make them compatible with Ivy
- Enables the `enableIvy` flag on the demo page
- Resolves Ivy compilation issues


projects/swimlane/ngx-ui/package.json
```diff
@@ -45,9 +45,9 @@
     "json-schema": "*"
   },
   "dependencies": {
-    "@angular/flex-layout": "8.0.0-beta.26",
+    "@angular/flex-layout": "8.0.0-beta.27",
```
duredhel13: Are we sure about this? I think it's safe to double check that the splitDirective is still working.

src/app/app.module.ts
```diff
@@ -5,10 +5,10 @@ import { FormsModule } from '@angular/forms';
 import { NgxDatatableModule } from '@swimlane/ngx-datatable';
 
 import { MomentModule } from 'ngx-moment';
-import { NgxUIModule } from '../../projects/swimlane/ngx-ui/src/public_api';
+import { NgxUIModule } from '@swimlane/ngx-ui';
```
duredhel13: Wouldn't this make it a bit harder to update the docs?

marjan-georgiev: It's still importing directly from source, I updated the paths. See here: https://github.com/swimlane/ngx-ui/pull/304/files#diff-e5e546dd2eb0351f813d63d1b39dbc48L17-L18

projects/swimlane/ngx-ui/package.json
marjan-georgiev: Good catch. Updated the demo pages. Looks like it doesn't like using the directive name as an input:

`<div [ngxSplit]="'row'">`

but likes this:

`<div ngxSplit [ngxSplit]="row">`

marjan-georgiev: But that's probably Ivy-related, not flex-layout necessarily.

PR #309, opened 2019-10-24
Change InjectionService to use portals

marjan-georgiev:
Updates the injection service to use CDK portals


luisb90:
gotta increase the version number / update changelog

marjan-georgiev:
> gotta increase the version number / update changelog

We only do that when we release. Not every PR is gonna have its own release.

PR #315, opened 2019-11-05
Json Editor Flat

duredhel13:
Adds a 2nd User Interface for the Json Editor. 
This does not include the advance schema editor.

usage: `<ngx-json-editor-flat ... ></ngx-json-editor-flat>`

image

image


projects/swimlane/ngx-ui/src/lib/components/json-editor/json-editor-flat/json-editor-node-flat/json-editor-node-flat.component.html
```diff
@@ -0,0 +1,125 @@
...
+          <span class="property-name">{{ schema?.propertyName | cammeltosnake }}</span>
+        </div>
+
+        <span class="description">{{ schema?.description }}</span>
+      </div>
+
+      <div class="node-input">
+        <!-- Number | Integer -->
+        <div *ngIf="schema?.type === 'number' || schema?.type === 'integer'">
+          <ngx-input type="number" [ngModel]="model" (ngModelChange)="updateModel(parseInt($event))"
```
marjan-georgiev: Can probably do better coercion, since this will convert floats to integers too, but that's a bigger undertaking for the whole library, so extracted it into a ticket: https://github.com/swimlane/ngx-ui/issues/321

Maybe change this to `Number($event')` for now? Don't know if that even works in templates.

projects/swimlane/ngx-ui/src/lib/components/json-editor/json-editor-flat/json-editor-node-flat/json-editor-node-flat.component.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  ngOnInit() {
+    this.level += 1;
+    if (this.level > 1) {
+      this.indentationArray = Array(this.level - 1).fill(this.level);
+    }
+  }
+
+  parseInt(value: string): number | null {
+    return value ? parseInt(value) : null;
```
marjan-georgiev: The linter complains on parseInt not being passed a radix param:

```suggestion
    return value ? parseInt(value, 10) : null;
```

duredhel13:
> Correct me if I'm wrong, but shouldn't these fields be red if they are 1: required with no value, or 2: a number field with an invalid input?
> 
> Screen Shot 2020-01-06 at 9 44 34 AM

@aacebo you're totally right. Looks like a bug on `ngx-input` though.

PR #318, opened 2019-11-14
Pass down a custom schema validator to the JSON Editor

duredhel13:


projects/swimlane/ngx-ui/src/lib/components/json-editor/json-editor.component.ts
```diff
@@ -31,6 +32,8 @@ export class JsonEditorComponent implements OnChanges {
   @Input()
   typeCheckOverrides?: any;
 
+  @Input() customAjv: Ajv.Ajv;
```
marjan-georgiev: Would prefer this to be a method, so it doesn't add a dependecy on Ajv. Allows users to use different schema validators.

PR #319, opened 2019-11-17
Json Editor Flat Schema Builder

duredhel13:
JSON Editor Schema Builder Mode based on the new `ngx-json-editor-flat` interface.

Designs:
- https://www.figma.com/file/fGr0PYeAiHLTDjqCaoIOI6/Plugin-IDE?node-id=206%3A0

Usage: 
```
jsonEditorSchemaBuilderModel: any = {};

customFormats = ['password', 'code', 'date', 'date-time', 'custom'];

<ngx-json-editor-flat
      [(model)]="jsonEditorSchemaBuilderModel"
      [schema]="{}"
      label="Schema"
      [typeCheckOverrides]="typeOverrides"
      [schemaBuilderMode]="true"
      [formats]="customFormats"
      (schemaChange)="schemaChange($event)">
 </ngx-json-editor-flat>
```

Couple of things missing from this implementation and will be added as separate PRs:
- The ability to add examples to an object property.
- The ability for an object property to have multiple types (currently only one is supported).


Hypercubed:
I belive the component, as-is, is not OnPush compatible and will not work well on pages that are OnPush.

projects/swimlane/ngx-ui/src/lib/components/json-editor/json-editor/json-editor-node/node-types/object-node/object-node.component.html
```diff
@@ -1,29 +1,29 @@
 <div [hidden]="!expanded">
-  <div *ngFor="let prop of propertyIndex | keyvalue; trackBy: trackBy">
+  <div *ngFor="let prop of getProperties(); trackBy: trackBy">
```
Hypercubed: I think the pipe was better here.... this is going to run on each "digest".

duredhel13: The pipe was messing with the order of properties. Was listing props by alphabetic order.

Hypercubed: I'd suggest... if possible... create your own pipe.  Would be better than repeatedly calling `getProperties`... which returns a new array each time.

duredhel13: Done

Issue #314, opened 2019-10-31
Disable keydown.enter on confirm dialog if longPress is true


marjan-georgiev:
Pressing enter on the longPress confirm dialog emits the "OK" event, which bypasses the long press button. It should be disabled if the longPress setting is true.

PR #323, opened 2019-12-02
disable keydown when longpress is true

haeminn:
Addressing #314


marjan-georgiev:
LGTM! Tests failing are unrelated, will be fixed in a different PR. Merging

Issue #321, opened 2019-11-19
Coerce inputs


marjan-georgiev:
Our input components do not convert the inputs into the expected value types (i.e. boolean, integer)
This issue is present in both our inputs, as well as the json-editor component.

Should use @angular/cdk/coercion package

Reference:
https://netbasal.com/trust-but-verify-coerce-your-component-inputs-bdb743e8b579

aacebo:
we are incrementally fixing this issue with feature branches, soon to be finished.

PR #324, opened 2019-12-02
Coerce inputs

haeminn:
Addresses #321


projects/swimlane/ngx-ui/src/lib/components/input/input.component.ts
```diff
@@ -24,7 +24,7 @@ import {
 } from '@angular/forms';
 import { BehaviorSubject } from 'rxjs';
 
-import { coerceBoolean } from '@swimlane/ngx-ui/utils';
```
haeminn: Looks like this function is exactly the same as `coerceBooleanProperty`. Should we remove this and replace with `coerceBooleanProperty` in other components that also use that?

aacebo: yes, going to open a PR to remove our internal util, this was only made because something was originally preventing us from referencing the cdk version.

projects/swimlane/ngx-ui/src/lib/components/input/input.component.ts
```diff
@@ -213,7 +213,7 @@ export class InputComponent implements AfterViewInit, ControlValueAccessor, Vali
   focused: boolean = false;
   readonly type$ = new BehaviorSubject<InputTypes>(undefined);
 
-  private _value: string = '';
+  private _value: any = '';
```
Hypercubed: Can this be `string | number`?

projects/swimlane/ngx-ui/src/lib/components/input/input.component.ts
```diff
@@ -168,10 +168,11 @@ export class InputComponent implements AfterViewInit, ControlValueAccessor, Vali
...
   @ViewChild('textareaControl', { static: false }) readonly textareaControl: ElementRef<HTMLTextAreaElement>;
 
-  get value(): string { return this._value; }
-  set value(val: string) {
+  get value(): string | number { return this._value; }
+  set value(val: string | number) {
     if (val !== this._value) {
-      this._value = val;
+      this._value = this.type === InputTypes.number ? coerceNumberProperty(val) : val;
+      console.log(typeof this._value)
```
marjan-georgiev: Console.log

PR #326, opened 2019-12-03
Luisb90/tabs on push

luisb90:
Updates tabs to use OnPush change detection method and adds unit tests to bring component coverage to 100%.

Screen Shot 2019-12-03 at 1 26 15 PM


luisb90:
> the only thing I noticed was that the `components/tabs-page` component wasn't set to on-push.

woops, updated

PR #327, opened 2019-12-03
remove coercion util and use cdk

aacebo:
- remove internal coercion util
- replace usage of util with angular cdk method


projects/swimlane/ngx-ui/src/lib/components/button/button.component.ts
```diff
@@ -1,7 +1,7 @@
 import { Component, Input, ViewEncapsulation, OnInit, OnChanges, HostListener, ChangeDetectionStrategy } from '@angular/core';
+import { coerceBooleanProperty } from '@angular/cdk/coercion';
```
luisb90: how nice of angular to include this :)

aacebo: hopefully this fixes that weird issue you were seeing when integrating with platform...

PR #332, opened 2019-12-06
Use border to indicate timeout for notifications

Hypercubed:
Preview: https://www.screencast.com/t/sHnKnE8qYJW


projects/swimlane/ngx-ui/src/lib/components/notification/notification.component.ts
```diff
@@ -35,12 +39,18 @@ export class NotificationComponent {
...
   @Input() icon: string;
+  @Input() timeout: false | number;
 
   @Output() close = new EventEmitter();
   @Output() pause = new EventEmitter();
   @Output() resume = new EventEmitter();
 
-  timeout: any;
+  get animationDuration() {
+    if (this.timeout === false) {
```
luisb90: do we want a falsy check here? or should we default `timeout` to false?

luisb90: ahh yeah, yeah let's go with that I think, so it handles the `undefined` case and lets 0 through

PR #334, opened 2019-12-09
Update toolbar to use OnPush change detection

luisb90:
Updated the toolbar to add OnPush change detection and added unit tests


projects/swimlane/ngx-ui/src/lib/components/toolbar/toolbar.component.ts
```diff
@@ -83,7 +48,8 @@ export class ToolbarComponent {
   }
   constructor(private renderer: Renderer2, private elRef: ElementRef) {}
 
-  onMenuClicked(item, $event) {
+  onMenuClicked(item: ToolbarMenuItem, $event: any) {
```
aacebo: can we change this event and its emitter to be type `Event`?

projects/swimlane/ngx-ui/src/lib/components/toolbar/fixtures/toolbar.fixture.ts
```diff
@@ -0,0 +1,40 @@
+import { Component, ViewChild, ChangeDetectionStrategy } from '@angular/core';
```
aacebo: I like that you broke this into its own file, will update my code to do the same.

projects/swimlane/ngx-ui/src/lib/components/toolbar/toolbar.component.ts
```diff
@@ -83,7 +48,8 @@ export class ToolbarComponent {
...
+    /* istanbul ignore else */
```
aacebo: we should either test this path and remove the ignore, or remove the if statement if `item.click` is not nullable

projects/swimlane/ngx-ui/src/lib/components/toolbar/toolbar.component.ts
```diff
@@ -83,7 +48,8 @@ export class ToolbarComponent {
   }
```
aacebo: I think you forgot to put a new line between the end curly and the constructor.

projects/swimlane/ngx-ui/src/lib/components/toolbar/toolbar.component.ts
luisb90: typed it up

projects/swimlane/ngx-ui/src/lib/components/toolbar/toolbar.component.ts
luisb90: I added a test thatll test if the click function was not defined for a menu item

projects/swimlane/ngx-ui/src/lib/components/toolbar/toolbar.component.ts
luisb90: woops, updated

PR #335, opened 2019-12-09
update tooltip component to use OnPush

luisb90:
updates tooltip component to use OnPush change detection. unit tests will be included in a separate pr.


projects/swimlane/ngx-ui/src/lib/components/tooltip/tooltip.component.ts
```diff
@@ -18,44 +19,25 @@ import { AlignmentTypes } from './alignment.type';
...
+  @Input() type: StyleTypes;
+  @Input() placement: PlacementTypes;
+  @Input() alignment: AlignmentTypes;
+  @Input() spacing: number;
+  @Input() cssClass: string;
+  @Input() title: string;
+  @Input() template: any;
+  @Input() context: any;
+
+  @ViewChild('caretElm', { static: true }) caretElm;
```
aacebo: This could be typed as an`ElementRef<HTMLSpanElement>`

projects/swimlane/ngx-ui/src/lib/components/tooltip/tooltip.directive.ts
```diff
@@ -20,7 +20,7 @@ import { TooltipContentComponent } from './tooltip.component';
 import { TooltipService } from './tooltip.service';
```
aacebo: TooltipContentComponent is an unused import here

projects/swimlane/ngx-ui/src/lib/components/tooltip/tooltip.component.ts
```diff
@@ -18,44 +19,25 @@ import { AlignmentTypes } from './alignment.type';
...
-  caretElm;
+  @Input() host: any;
+  @Input() showCaret: boolean;
+  @Input() type: StyleTypes;
+  @Input() placement: PlacementTypes;
+  @Input() alignment: AlignmentTypes;
+  @Input() spacing: number;
+  @Input() cssClass: string;
+  @Input() title: string;
+  @Input() template: any;
```
aacebo: could this be typed as a TemplateRef? Also in the directive.

projects/swimlane/ngx-ui/src/lib/components/tooltip/tooltip.directive.ts
luisb90: it's actually needed by the `InjectionRegisteryService` it's inheriting from, but i added a type to it instead of any

projects/swimlane/ngx-ui/src/lib/components/tooltip/tooltip.component.ts
luisb90: updated

projects/swimlane/ngx-ui/src/lib/components/tooltip/tooltip.component.ts
luisb90: updated

PR #337, opened 2019-12-10
Luisb90/slider onpush

luisb90:
update slider component to be OnPush and add unit tests


projects/swimlane/ngx-ui/src/lib/components/slider/slider.component.ts
```diff
@@ -25,43 +26,14 @@ const edge = window.navigator.userAgent.indexOf('Edge') > -1;
 
 @Component({
```
aacebo: can we try to add `exportAs: 'ngxSlider'` when possible to our components, it allows for syntax like so which gives to component instance as a variable in the markup
```
<ngx-slider
  #slider="ngxSlider"
></ngx-slider>
```

luisb90: good call, added

PR #340, opened 2019-12-16
Luisb90/toggle onpush

luisb90:
update toggle component to use OnPush change detection strategy and add unit tests

### DIFF
--- a/projects/swimlane/ngx-ui/src/lib/components/toggle/toggle.component.ts
+++ b/projects/swimlane/ngx-ui/src/lib/components/toggle/toggle.component.ts
@@ -1,7 +1,6 @@
 import {
   Component,
   EventEmitter,
-  HostBinding,
   Input,
   Output,
   ViewEncapsulation,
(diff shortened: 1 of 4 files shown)


projects/swimlane/ngx-ui/src/lib/components/toggle/toggle.component.ts
```diff
@@ -11,45 +21,24 @@ let nextId = 0;
...
-  disabled: boolean = false;
-  @Input()
-  required: boolean = false;
-  @Input()
-  tabIndex: number = 0;
-  @Input()
-  label: string;
+  @Input() id: string = `toggle-${++nextId}`;
+  @Input() name: string = null;
+  @Input() disabled: boolean = false;
```
aacebo: can we use coerceBoolean and coerceNumber with getters/setters here so we don't have to bind for static inputs?

aacebo: also, has the demo page been set to on-push for this? I didn't see it in the changes.

luisb90: woops, missed the demo page, updated that and added coercion to the inputs

PR #342, opened 2019-12-17
Luisb90/drawer onpush

luisb90:
update drawer component to use OnPush change detection and use type coercion on inputs

### DIFF
--- a/projects/swimlane/ngx-ui/src/lib/components/drawer/drawer.component.ts
+++ b/projects/swimlane/ngx-ui/src/lib/components/drawer/drawer.component.ts
@@ -7,11 +7,13 @@ import {
   HostListener,
   ViewEncapsulation,
   OnDestroy,
-  ChangeDetectionStrategy
+  ChangeDetectionStrategy,
+  TemplateRef
 } from '@angular/core';
 import { trigger } from '@angular/animations';
 
 import { drawerTransition } from '../../animations/animations';
+import { coerceNumberProperty, coerceBooleanProperty } from '@angular/cdk/coercion';
 
 @Component({
   selector: 'ngx-drawer',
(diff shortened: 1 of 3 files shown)


projects/swimlane/ngx-ui/src/lib/components/drawer/drawer.component.ts
```diff
@@ -7,19 +7,25 @@ import {
...
   ViewEncapsulation,
   OnDestroy,
-  ChangeDetectionStrategy
+  ChangeDetectionStrategy,
+  TemplateRef
 } from '@angular/core';
 import { trigger } from '@angular/animations';
 
 import { drawerTransition } from '../../animations/animations';
+import { coerceNumberProperty, coerceBooleanProperty } from '@angular/cdk/coercion';
```
aacebo: these imports probably belong up with the other angular imports

projects/swimlane/ngx-ui/src/lib/components/drawer/drawer.component.ts
```diff
@@ -111,23 +126,20 @@ export class DrawerComponent implements OnDestroy {
...
-  @HostBinding('style.transform')
   transform: string;
 
   /**
    * Drawer width calculation
    *
    * @memberOf DrawerComponent
    */
-  @HostBinding('style.width')
   widthSize: any;
```
aacebo: this as well as `heightSize` could probably be either `string` or `string | number`

luisb90: yeah good call, updated

projects/swimlane/ngx-ui/src/lib/components/drawer/drawer.component.ts
luisb90: moved it up

PR #343, opened 2019-12-17
Aacebo/file upload button  on push

aacebo:
Summary:
- change detection on-push
- move markup to html files
- cleanup markup
- input coercion and typing added

### DIFF
--- a/projects/swimlane/ngx-ui/src/lib/components/button/file-button.component.ts
+++ b/projects/swimlane/ngx-ui/src/lib/components/button/file-button.component.ts
@@ -11,7 +11,8 @@ import {
   ViewChild,
   ElementRef
 } from '@angular/core';
-import { FileUploaderOptions, FileUploader } from '@swimlane/ng2-file-upload';
+import { coerceBooleanProperty } from '@angular/cdk/coercion';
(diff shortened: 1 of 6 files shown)


projects/swimlane/ngx-ui/src/lib/components/button/file-button.component.ts
```diff
@@ -111,20 +69,24 @@ export class FileButtonComponent implements OnInit {
...
-      success: this.isItemSuccessful,
+      success: this._isItemSuccessful,
       active: this.uploader && this.uploader.isUploading
     };
   }
 
-  isItemSuccessful: boolean = false;
-  progress: string = '0%';
+  readonly FileButtonStyleType = FileButtonStyleType;
+  progress: number = 0;
```
luisb90: not too big a deal since we're under construction but this would be a breaking change right? if someone consuming it was expecting the emitter to emit a string

aacebo: Ya it could, but we are the lib and there are going to be times when the inputs/outputs might need to change for components, we just have to be aware of it when we integrate it into platform and turbine frontends.

luisb90: yup, and i like this change as well. ship it 🚀

PR #351, opened 2019-12-21
Aacebo/drawer  on push

aacebo:
Summary:
- remove doc comments that are inconsistent with the rest of the project
- add types
- unit test


projects/swimlane/ngx-ui/src/lib/components/drawer/drawer.component.spec.ts
```diff
@@ -17,10 +18,80 @@ describe('DrawerComponent', () => {
...
+      expect(spy).toHaveBeenCalled();
+    });
+  });
+
+  describe('setDimensions', () => {
+    it('should set size when direction left', done => {
+      component.direction = DrawerDirection.Left;
+      component.setDimensions(10);
+
+      setTimeout(() => {
```
luisb90: why a timeout here? would `fixture.detectChanges` and `fixture.whenStable` work better here?

PR #368, opened 2020-01-21
In-component drawer

luisb90:
This PR updates the drawer component to give it the ability to open a drawer from within a parent container. an `isRoot` and `parentContainer` options have been added to the drawer options. When `isRoot` is set to false, you can supply an html element as a parent container, and the injection service will use the Portal api to inject the drawer as a child of that parent element. As part of this PR i also cleaned up code that wasnt being used properly.

Screen Shot 2020-01-21 at 7 35 29 AM

Screen Shot 2020-01-21 at 7 36 31 AM


projects/swimlane/ngx-ui/src/lib/components/drawer/drawer.component.ts
```diff
@@ -90,48 +93,18 @@ export class DrawerComponent implements OnDestroy {
...
-      } else {
-        transform = 'translate(0, 100%)';
-      }
-    }
-
-    setTimeout(() => {
-      this.heightSize = height;
-      this.widthSize = width;
-      this.transform = transform;
-    });
```
luisb90: I removed this code because during testing, I noticed that this code that was calculating the transform was never updating the element's transform property. The transform set by the drawer animation always took precedence and would not change even after subsequent calls to `setDimensions`. I also noticed that the code was calculating exact px sizes, even though the size property of the element is in percentages relative to 100%. It worked in that way when the width or height were just set to the specified percentage value. So I was able to greatly simplify this code.

projects/swimlane/ngx-ui/src/lib/components/drawer/drawer.component.ts
```diff
@@ -81,7 +80,7 @@ export class DrawerComponent implements OnInit, OnDestroy {
   transform: string;
   widthSize: string | number;
   heightSize: string | number;
-  position: 'fixed' | 'initial' = 'fixed';
+  position: 'fixed' | 'absolute' = 'absolute';
```
aacebo: can we change this to an enum instead of using magic strings?

src/app/dialogs/drawer-page/drawer-page.component.html
```diff
@@ -162,8 +162,8 @@ <h1>Attack Type: Malware</h1>
   </ngx-section>
 
   <ngx-section class="shadow" sectionTitle="Adding drawer to parent component instead of root">
-    <button type="button" (click)="openDrawer('bottom', undefined, false, undefined, false, parentContainer)">open drawer time</button>
+    <button type="button" (click)="openDrawer('left', undefined, true, editTmpl2, false, parentContainer)">open drawer time</button>
       <div [style.width.%]="100" [style.height.px]="500" [style.background-color]="'red'">
-        <div [style.height.%]="100" #parentContainer></div>
+        <div [style.height.%]="100" [style.position]="'relative'" #parentContainer></div>
```
aacebo: why are we using variable binding syntax with static values here? Instead of doing this you can create classes in the scss

aacebo:
If you go to your in-component demo and open the drawer, then click 'open details', the new global drawer opens but a backdrop is over the entire screen that renders all other elements unclickable.

src/app/dialogs/drawer-page/drawer-page.component.html
luisb90: none of the pages have scss files but you right, doesnt mean cant take 15 secs to add one ;) added and moved styles

projects/swimlane/ngx-ui/src/lib/components/drawer/drawer.component.ts
luisb90: enumed it up

luisb90:
> If you go to your in-component demo and open the drawer, then click 'open details', the new global drawer opens but a backdrop is over the entire screen that renders all other elements unclickable.

Fixed, sneaky typo in the host bindings

luisb90:
> Getting the following error when I trigger a notification on the notifications demo page:
> 
> image
> 
> Assuming caused by the change in the injection service

updated and fixed, all dialog components should be working as expected

PR #372, opened 2020-01-24
#372: fix button styling on .ngx-tabs

hankconn:
Tested this locally and it's working for me.


marjan-georgiev:
@hankconn looks like prettier is still not happy. Try running `npm run prettier` to format all files.

PR #375, opened 2020-01-27
Aacebo/bug  drawer resize

aacebo:
Summary:
- resize for root level drawers only happens on init
- resize on window resize when root
- add unit tests
- fix markup tabbing

### DIFF
--- a/projects/swimlane/ngx-ui/src/lib/components/drawer/drawer.component.ts
+++ b/projects/swimlane/ngx-ui/src/lib/components/drawer/drawer.component.ts
@@ -110,4 +110,11 @@ export class DrawerComponent implements OnInit, OnDestroy {
   onEscapeKey(): void {
     this.close.emit(true);
   }
+
+  @HostListener('window:resize')
+  onWindowResize(): void {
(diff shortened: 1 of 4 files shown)


projects/swimlane/ngx-ui/src/lib/components/drawer/drawer.component.ts
```diff
@@ -110,4 +110,11 @@ export class DrawerComponent implements OnInit, OnDestroy {
   onEscapeKey(): void {
     this.close.emit(true);
   }
+
+  @HostListener('window:resize')
+  onWindowResize(): void {
```
marjan-georgiev: Should we throttle this?

aacebo: ya, we could use a set timeout there just would be a tiny delay on the resize

aacebo: after looking into it I am unable to reproduce the issue anymore and am closing this PR

Issue #262, opened 2019-03-13
Use json-schema examples as placeholders in json-editor


marjan-georgiev:
If a json-schema type has an example property, use those in as placeholders in the inputs in the json-editor component

aacebo:
my recent PR fixes this for the original json-editor and the new flat version keep the examples in a tooltip so closing

PR #376, opened 2020-01-28
Aacebo/json editor  examples placeholder

aacebo:
Summary:
- when examples array is defined in schema it should be used as placeholder for node if possible

issue: https://github.com/swimlane/ngx-ui/issues/262

### DIFF
--- a/projects/swimlane/ngx-ui/src/lib/components/json-editor/json-editor-flat/json-editor-node-flat/json-editor-node-flat.component.html
+++ b/projects/swimlane/ngx-ui/src/lib/components/json-editor/json-editor-flat/json-editor-node-flat/json-editor-node-flat.component.html
@@ -29,7 +29,7 @@
       <div class="node-input">
         <!-- Number | Integer -->
         <div *ngIf="schema?.type === 'number' || schema?.type === 'integer'">
-          <ngx-input type="number" [placeholder]="placeholder" [ngModel]="model" (ngModelChange)="updateModel($event)" [requiredIndicator]="false"
+          <ngx-input type="number" [ngModel]="model" (ngModelChange)="updateModel($event)" [requiredIndicator]="false"
             [required]="required"></ngx-input>
         </div>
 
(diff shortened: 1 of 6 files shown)


projects/swimlane/ngx-ui/src/lib/components/json-editor/json-editor-flat/json-editor-node-flat/json-editor-node-flat.component.html
```diff
@@ -44,11 +44,11 @@
         <ng-container *ngIf="schema?.type === 'string'">
           <!-- No format -->
           <div *ngIf="!schema.format">
-            <ngx-input *ngIf="!schema?.enum" type="text" [ngModel]="model" (change)="updateModel($event)"
+            <ngx-input *ngIf="!schema?.enum" type="text" [placeholder]="placeholder" [ngModel]="model" (change)="updateModel($event)"
               [requiredIndicator]="false" [required]="required">
             </ngx-input>
 
-            <ngx-select [filterable]="false" *ngIf="schema?.enum" [ngModel]="[model]"
+            <ngx-select [filterable]="false" [placeholder]="placeholder" *ngIf="schema?.enum" [ngModel]="[model]"
```
duredhel13: Hmm, not sure if it makes sense here since the user is already selecting from a predefined list of values.

projects/swimlane/ngx-ui/src/lib/components/json-editor/json-editor/json-editor-node/json-editor-node.component.ts
```diff
@@ -18,7 +18,17 @@ export class JsonEditorNodeComponent extends JsonEditorNode {
...
+  placeholder: string = '';
+
   constructor(public dialogMngr: DialogService) {
     super(dialogMngr);
   }
+
+  ngOnInit() {
+    super.ngOnInit();
+
+    if (this.schema.examples && Array.isArray(this.schema.examples)) {
```
duredhel13: Can we have this on the base class?

aacebo: good call

projects/swimlane/ngx-ui/src/lib/components/json-editor/json-editor-flat/json-editor-node-flat/json-editor-node-flat.component.html
aacebo: can the select ever be in a state where nothing is selected? Thats when the placeholder would come in handy so it should matter if they are selecting from a list just if the value can be undefined.

aacebo: oh I see we haven't implemented placeholders within ngx-select

duredhel13: I was going to say to leave the placeholder. But it’s not supported then?

projects/swimlane/ngx-ui/src/lib/components/json-editor/json-editor-flat/json-editor-node-flat/json-editor-node-flat.component.html
```diff
@@ -29,7 +29,7 @@
       <div class="node-input">
         <!-- Number | Integer -->
         <div *ngIf="schema?.type === 'number' || schema?.type === 'integer'">
-          <ngx-input type="number" [ngModel]="model" (ngModelChange)="updateModel($event)" [requiredIndicator]="false"
+          <ngx-input type="number" [placeholder]="placeholder" [ngModel]="model" (ngModelChange)="updateModel($event)" [requiredIndicator]="false"
```
duredhel13: As per our discussion, we don’t need to do this in the flat version. We will show examples in a tooltip.

PR #377, opened 2020-01-29
Add selectable and disabled to tree node model

rhalff:
It seems `selectable` and `disabled` are not configurable for ngx-tree when no templates are used.

As a result it's not possible to disable or make anything selectable with the stock configuration of ngx tree.

In addition I've added `selectNode` to the nested trees in order for that event to 'bubble' up and I'm canceling propagation when a node is expanded otherwise an expand will also trigger a select which seems undesirable.

This pr purposes a suggested change and does not have to be merged perse.


marjan-georgiev:
Looks good, thanks!

PR #380, opened 2020-02-03
Luisb90/ngx stepper

aacebo:
Summary:
- create ngx-stepper feature
- create demo page
- use angular animations to enhance user experience
- re-work original design to work with future use-cases outlined in figma
- pull in resize-observer polyfill/typings package to create `resizeObserver` directive that allows us to observe dom element size changes in an event driven way
- unit test new features


projects/swimlane/ngx-ui/src/lib/components/stepper/step.component.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  private _active?: number;
+  private _step?: number;
+  private _total?: number;
+  private _completeIcon?: string;
+
+  constructor(private readonly _cdr: ChangeDetectorRef, private readonly _el: ElementRef<HTMLElement>) {}
+
+  ngOnInit() {
+    this._cdr.markForCheck();
```
luisb90: all of the input properties are marking for check as well, when they were init you still needed an extra mark for check?

projects/swimlane/ngx-ui/src/lib/components/stepper/stepper.animation.ts
```diff
@@ -0,0 +1,74 @@
...
+      { params: { width: 0 } }
+    ),
+    state(
+      StepperBarAnimationStates.Move,
+      style({
+        transition: '500ms cubic-bezier(0.35, 0, 0.25, 1)',
+        transform: 'translateX({{ width }}px)',
+        height: `${BAR_THICKNESS}px`,
+        width: `${BAR_SIZE / 2}px`,
+        margin: `0 ${BAR_SIZE / 4}px 0 ${BAR_SIZE / 4 - (BAR_SIZE - STEP_SIZE) / 2}px`
```
luisb90: [MATH INTENSIFIES]

projects/swimlane/ngx-ui/src/lib/components/stepper/stepper.component.ts
```diff
@@ -0,0 +1,203 @@
...
+  @ContentChildren(StepComponent)
+  get steps() {
+    return this._steps;
+  }
+  set steps(v) {
+    this._steps = v;
+    this._destroy$.next();
+
+    for (const item of this._steps.map((step, i) => ({ step, i }))) {
+      setTimeout(() => {
```
luisb90: so even though all these properties have mark for check, you still needed a setTimeout?

projects/swimlane/ngx-ui/src/lib/components/stepper/step.component.ts
aacebo: yes, because the inputs can change dynamically after the component has been initialized, but inputs are normally set prior to component initialization on load so its best to hit both to cover all the bases.

projects/swimlane/ngx-ui/src/lib/components/stepper/stepper.component.ts
aacebo: Yes, if not you will get weird change detection behavior, this is so that we can make all changes outside angular then do a check after the fact. Not ideal but the only performant solution I could find.

luisb90: figured something like that was the reason, just wanted to double check 👍

PR #384, opened 2020-02-05
Add showEvent to dropdown component

Hypercubed:


projects/swimlane/ngx-ui/src/lib/components/dropdown/dropdown-toggle.directive.ts
```diff
@@ -1,6 +1,12 @@
 import { Directive, Input, ElementRef, HostListener, EventEmitter, Output } from '@angular/core';
 import { coerceBooleanProperty } from '@angular/cdk/coercion';
 
+export enum DropdownShowTypes {
```
aacebo: this should be in its own file

src/app/components/dropdown-page/dropdown-page.component.html
```diff
@@ -247,4 +247,57 @@ <h1 style="padding: 10px;">Hello!</h1>
       </ngx-dropdown>
       ]]>
   </ngx-codemirror>
+</ngx-section>
+
+<ngx-section class="shadow" [sectionTitle]="'Show Events'">
```
aacebo: when using static inputs in all the areas below you do not need to bind it using `[input]="'value'"` you can just use `input="value"` since you know it isn't going to change

aacebo: we have plans to update the demo pages, just haven't gotten to it yet.

Hypercubed: I think for the code samples we should keep the square brackets... to communicate that this is a component input.  Elsewhere we can switch.

aacebo: We are going to be adding an API tab to communicate inputs/outputs to the users.

PR #386, opened 2020-02-06
Export schema on model update

duredhel13:
Expose a schema update when model changes, without modifying the original schema.
Updates on add property are already on master. This PR fixes the delete property.


duredhel13:
> Should we add this to the json-editor component too?

@marjan-georgiev not sure if we need to. There will be some extra work/refactor to make it work there.

Issue #373, opened 2020-01-27
Slider component needs initial value for 'background-size' property in case of filled=true


kivinm:
Steps to reproduce:
Use a slider component with [filled]="true" and initial value = 0.

Expected result:
Slider fill-bar is empty.

Actual result:
Slider fill-bar is 100% filled.

The reason for this bug is that you don't set initial value for 'background-size' property of fill-bar.
When initial value = 0, setValues() method and code inside it (which sets fill-bar style) isn't called.

Solution: provide initial value for 'background-size' property in slider.component.scss

&.filled .fill-bar {
    background-size: 0% 100%;
    ...

PR #392, opened 2020-02-11
fix initial value = 0 bug on slider component

haeminn:
#373


Hypercubed:
I think, maybe instead of unsetting the default vaue, you shoudl. be able to call `this.setValues(0);` in ngOnInit.

PR #397, opened 2020-02-19
Enhancement: delay closeOnMouseLeave in dropdown

Hypercubed:
Added 1s delay on mouseleave event for dropdown to prevent closing in incedental mouseleave.


marjan-georgiev:
Prettier is complaining. `npm run prettier` to fix.

PR #398, opened 2020-02-19
Move click event on tree nodes to avoid propagation

Hypercubed:
* Moved click event on node to avoid using stop propagation in click event.


src/app/components/tree-page/tree-page.component.ts
```diff
@@ -88,4 +90,8 @@ export class TreePageComponent {
     "children": [],
     "spouse": null
   }`);
+
+  selectNode(e: Event) {
+    console.log({ e });
```
haeminn: Do we need to log the event?

Hypercubed: Yeah... think it's good here for testing.

PR #437, opened 2020-04-30
Luisb90/ngx navbar

luisb90:
## Summary

Added a navbar component with configurable nav items. An animated sliding bar indicates which nav item is currently active. Added a new demo page to showcase the component.

## Checklist

- [x] \*Added unit tests
- [x] \*Added a code reviewer
- [x] Updated the demo page

_\*required_

Screen Shot 2020-04-30 at 10 18 51 AM


luisb90:
> Code and tests look good but there is some change detection funny business happening when the component is initialized to an index > 0

added a routing example to the demo page and fixed up the init behavior. also ran latest prettier

luisb90:
> Looks good, a few comments/questions:
> 
> 1. Can we add code examples in the demo page?
> 2. Our left nav component is called `ngx-nav-menu`, and this is called `ngx-nav`. Should we consider giving them more distinguishable names now that we have two different types of navs?
> 3. Is the height correct? The icon doesn't seem vertically centered, and the component seems taller than what  our toolbars are.
> 
> image

1. Updated the demo page with code examples

2. Well the top level component is called `ngx-navbar`. I renamed the subcomponent from `ngx-nav` to `ngx-navbar-item`, is that better?

3. good catch, updated the height to match the height of the navbar in platform

Screen Shot 2020-05-05 at 12 06 44 AM
@marjan-georgiev

PR #446, opened 2020-05-27
Add isNotEmpty keyword to default SchemaValidator

Hypercubed:
## Summary

- Added `isNotEmpty` validation to schema validator

## Checklist

- [ ] \*Added unit tests
- [x] \*Added a code reviewer
- [x] Updated the demo page

_\*required_

### DIFF
--- a/projects/swimlane/ngx-ui/src/lib/components/json-editor/schema-validator.service.ts
+++ b/projects/swimlane/ngx-ui/src/lib/components/json-editor/schema-validator.service.ts
@@ -11,6 +11,24 @@ export class SchemaValidatorService {
     });
     this.ajv.addFormat('password', '.*');
     this.ajv.addFormat('code', '.*');
+
+    this.ajv.addKeyword('isNotEmpty', {
+      type: 'string',
+      validate: function myValidation(_schema: any, data: any) {
+        if ((myValidation as any).errors === null) (myValidation as any).errors = [];
+
+        if (typeof data === 'string' && data.trim() === '') {
+          (myValidation as any).errors.push({
+            keyword: 'isNotEmpty',
+            message: 'required',
+            params: { keyword: 'isNotEmpty' }
+          });
+          return false;
+        }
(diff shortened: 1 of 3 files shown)


src/app/components/json-editor-page/json-editor-page.component.ts
```diff
@@ -23,7 +23,8 @@ export class JsonEditorPageComponent {
       productName: {
         description: 'Name of the product',
         type: 'string',
-        examples: ['Apples', 'Oranges']
+        examples: ['Apples', 'Oranges'],
+        isNotEmpty: true
```
duredhel13, line 27: Hmm, then every property need to have this defined in the schema to show the validation error?

Hypercubed: Yes... for now.  Unless we decide that require is always `isNotEmpty`.

projects/swimlane/ngx-ui/src/lib/components/json-editor/schema-validator.service.ts
```diff
@@ -11,6 +11,24 @@ export class SchemaValidatorService {
     });
     this.ajv.addFormat('password', '.*');
     this.ajv.addFormat('code', '.*');
+
+    this.ajv.addKeyword('isNotEmpty', {
```
shaunburdick, line 15: We shouldn't break from the already defined specification like this. This could end up breaking our JSON Schema validation in other projects.
I'd prefer we find an alternative (something like [minLength](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.6.3.2) or creating an additional format which is allowed in the spec)

Hypercubed: We'll close this PR... shemas will need to specify the minLength.

But does this really break JSON Schema validation?  Doesn't it support adding new validation keywords?

PR #476, opened 2020-07-22
SPT-6967 - select input fill variant

luisb90:
## Summary

Updated the select component to include a "fill" appearance variant, as shown in the screenshot. This also included an updating of the dropdown styling that applies to both the fill and legacy variants, based on latest designs from UX. For the fill variant, the select will automatically fit to its contents up to the max width of its parent container. If no choice is selected, the select will initialize to the width of either the placeholder or the label if defined.

## Checklist

- [x] \*Added unit tests
- [x] \*Added a code reviewer
- [x] Updated the demo page

_\*required_

Screen Shot 2020-07-22 at 9 32 44 AM
Screen Shot 2020-07-22 at 9 33 42 AM
Screen Shot 2020-07-22 at 9 33 30 AM
Screen Shot 2020-07-22 at 9 33 20 AM
Screen Shot 2020-07-22 at 9 33 05 AM
Screen Shot 2020-07-22 at 9 33 00 AM

### DIFF
--- a/projects/swimlane/ngx-ui/CHANGELOG.md
+++ b/projects/swimlane/ngx-ui/CHANGELOG.md
@@ -9,6 +9,8 @@
 - Feature: sm, md, lg input sizing
 - Bug: make context drawer compatible with components
 - Bug: fix ngx-drawer-content scrolling scss issue
+- Feature: fill select style
+- Enhancement: select dropdown styling
 
 ## 29.1.0 (2020-06-29)
 
--- a/projects/swimlane/ngx-ui/src/lib/components/select/select-dropdown.component.html
+++ b/projects/swimlane/ngx-ui/src/lib/components/select/select-dropdown.component.html
@@ -1,5 +1,6 @@
 <div>
   <div class="ngx-select-filter" *ngIf="filterable && !tagging">
+    <i class="ngx-icon ngx-search"></i>
     <input
       #filterInput
       type="search"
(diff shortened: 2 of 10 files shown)


luisb90:
> 1. Please update `::placeholder` styles for the dropdown search, to `$color-blue-grey-350`
> 2. Please update `.ngx-select-empty-placeholder-text` to be `$color-blue-grey-150`
> 3. Legacy styles are incorrectly inheriting left padding on `.ngx-select-input-list`
> 4. The new auto-width receives the blue label treatment on focus, but not legacy styles. Legacy should be treated the same—see Single Select/Basic on the selects demo. Same applies to `ngx-select-caret` and `ngx-select-clear`.
> 5. .icon-arrow-down (.ngx-select-caret) and .icon-x (.ngx-select-clear) for both Legacy and Fill variants should be `$color-blue-grey-350` when blurred.
> 6. The designs may not have been overly clear—on the multi-select and tagging "tags", the X should be `$color-blue-grey-300` (always; regardless of field's focus) and `$color-blue-400` on hover of the icon itself.
> 7. Please add `line-height: 1.8` to `.class="ngx-select-input-option` and remove the explicit height (or tell me why it's there)
> 8. `.ngx-select-input-option` has left margin that creates an odd offset—can we do `margin: 2px 5px 2px 0;` on each option?
> 9. In conjunction with adjusting the option margin, could we change `.ngx-select-input-list` to have ` padding: 1px 40px 1px 10px;`
> 10. `.ngx-select-input-box-wrapper input` should match base input styles with `color: $color-blue-grey-100; font-size: 16px; line-height: 20px;`
> 11. Auto-width seems to be tied to the fill variant, auto-width should be compatible with legacy appearance as well.

@mcwenger all of these have been addressed in the last commit, you're good to rereview

PR #498, opened 2020-09-07
Issue 494 ngx select

nelsonGuti:
## Summary

Fix issue 494
Ngx-select bug where when using the filter and the arrow keys on the options dropdown, the `focusIndex` will update properly but would select an incorrect item because it would return the option on the focusIndex position of the total options array, and not the filtered array currently displayed. Fixed by passing the option to the `onOptionKeyDown` the user is on just as it is done on the click event.

## Checklist

- [ ] \*Added unit tests
- [x] \*Added a code reviewer
- [ ] Updated the demo page

_\*required_


luisb90 (approved):
looks good 👍

PR #505, opened 2020-10-05
performance improvements for ngx-ui

luisb90:
## Summary

The aim of this pr is to decrease overall page load times and improve responsiveness for the ngx-ui demo site. At the end of this pr I was able to decrease the overall gzipped bundle size from 1.01Mb to 545.18Kb, a 46.02% improvement. The main bundle size decreased from 978.86kb to 388.68Kb, a 60.3% improvement. Below are the steps I took and the gains along the way.

## 1. Clean up the npm security warnings

Initially, we have 16 vulnerabilities reported by npm:
```
found 16 vulnerabilities (15 low, 1 moderate) in 2268 scanned packages
  run `npm audit fix` to fix 13 of them.
  1 vulnerability requires semver-major dependency updates.
  2 vulnerabilities require manual review. See the full report for details.
```

I ran `npm audit fix` to take care of those. Only one remains now, part of the chokidar package:

```
found 1 low severity vulnerability
  run `npm audit fix` to fix them, or `npm audit` for details
```

## 2. Analyzing the initial bundle size

In order to get stats on how are bundle was being built and what was taking up the most space, I installed the `webpack-bundle-analyzer` package. I added the `--statsJson=true` flag to the `ng build` command in our `build:demo` script. This generates a file containing stats about the bundle. using `webpack-bundle-analyzer dist/ngx-ui/stats.json`, we can analyze the bundle in a visual way:

Screen Shot 2020-10-05 at 9 58 37 AM
Screen Shot 2020-10-05 at 8 46 03 AM

Looking at this, we can see things jumping out at us:

- The `faker` library has made it into our bundle. This library is fairly large and should only be used in development for stuff like unit tests.
- The `codemirror` library takes second in size. We should try and use this lightly, yet we're using it in readonly mode everywhere to display syntax highlighted code snippets. This is probably causing a performance hit.
- The `moment` library is like 80% locale files. We currently only need the `en` locale file, so it'd be an easy win to remove those
-  `moment-timezone` library similarly takes up space because of a lot of data files for different countries that we can shed.
- The `icons.json` file is a giant json object that contains svg data for all of our icons. We only pull it in to reference the icon names. We can probably create a name map and avoid pulling in this very large file.
- We reference the library code directly in our prod build, instead of the dist bundle created by building the library. If we switched to that, we could make use of any build optimizations used to create that bundle.
- Our main source code is gigantic. All of our demo pages are bundled into the main bundle, so we have to load everything just to visit one page. We can greatly improve this by implementing lazy loading.

## 3. Removing faker

faker's gzipped bundle size is 325.04Kb. It's huge. And we are currently only using it in the stepper page to generate random step labels, and in the select page to generate random option labels for one section. These are trivial uses and not worth the massive bundle size increase. I moved the library to `devDependencies`, and we only use it for our library unit tests now.

## 4. Removing readonly `ngx-codemirror` instances

Codemirror is nice in that it gives us a syntax highlighted code-editor environment that we insert anywhere in a page. But because it's so feature-rich, using it comes at a performance cost. We currently use it in readonly mode to display code snippets with syntax highlighting: 

Screen Shot 2020-10-05 at 10 41 45 AM

Syntax highlighting can be done without having to use codemirror with a lightweight library dedicated to this. I went with `Prism`. I created a wrapper component around the prism syntax `<pre><code class="language-{lang}"></code></pre>`, so that you only have to provide the CDATA of the code you want to highlight. Since it's within a `pre` tag, all line breaks and spaces are preserved. Here's an example:

```
        <app-prism>
<![CDATA[<button type="button" class="btn" (click)="open(DrawerDirection.Left)">Open Left</button>
<button type="button" class="btn" (click)="open(DrawerDirection.Bottom)">Open Bottom</button>]]>
        </app-prism>
```

I created a custom css theme to try and match our codemirror styles. The result of the same block above, done with Prism:

Screen Shot 2020-10-05 at 10 42 16 AM

Almost identical!

## 5. Optimizing `moment` and `moment-timezone`

moment is a useful but very large library for dealing with dates. A lot of the size comes from its need to be able to deal with different locales and timezones. Luckily, for our use case we only need the `en` locale. To help strip out all of the excess data, I installed the following packages:

- `@angular-builders/custom-webpack` - This lets us extends the webpack config `angular.json` uses under the hood
- `moment-locales-webpack-plugin` - a webpack plugin that lets us specify only the locale files we want for moment
- `moment-timezone-data-webpack-plugin` - a webpack plugin that lets us specify only the timezone data we need

To hook these up, we create a file at the root called `extra-webpack.config.js`. in `angular.json` we change the builder to reference the custom builder `"builder": "@angular-builders/custom-webpack:browser",`, and pass the following to options to tell it to reference the extra config file:

```
"customWebpackConfig": {
  "path": "./extra-webpack.config.js"
},
```

Configuring the moment plugins is simple afterwards. this is the entirety of the config file:

```
const MomentLocalesPlugin = require('moment-locales-webpack-plugin');
const MomentTimezoneDataPlugin = require('moment-timezone-data-webpack-plugin');
const currentYear = new Date().getFullYear();
 
module.exports = {
    plugins: [
        new MomentLocalesPlugin(),
        new MomentTimezoneDataPlugin({
          startYear: currentYear - 5,
          endYear: currentYear + 5,
      }),
    ],
};
```

Doing this, bundle size is substantially decreased, as we have removed all of the moment bloat.

## 6. Removing the need to import icons.json

The icons.json file contains information on all of our icons, included svg values. it comes in a 70Kb gzipped. We only use it to reference the name in the icons page, to render the icons using `ngx-icon` and label them. I created an array that serves as a name map instead:

```
// add new icon names to the end of this array
export const iconNameMap = [
  '3d-rotate',
  'add-circle-filled',
  'add-circle',
  'add-edge',
  'add-new',
  'add-node',
  'advanced-pie',
  ...
];
```
and export this. There is the added overhead of having to make sure you add the new name to this file as well whenever you need to add a new icon, but this removes the need to pull in `icons.json` into the icons page, making it only 1.37Kb in size.

## 7. Referencing the built library for prod builds

When we build our library, angular creates an optimized bundle that is minified. We should refer to this bundle when building the demo page for prod. To do this, I added `tsconfig.prod.json` which points the builder to the location of the built library:

```
{
  "extends": "./tsconfig.app.json",
  "compilerOptions": {
    "paths": {
      "@swimlane/ngx-ui": ["dist/swimlane/ngx-ui"],
      "@swimlane/ngx-ui/*": ["dist/swimlane/ngx-ui/*"],
      "tslib": ["../node_modules/tslib"]
    }
  }
}
```

The build library is about 100Kb smaller than packing in the code directly. For development we still point at the library source code, so any changes made are still immediately reflected in the demo page.

## 7. Splitting our pages into lazy modules

We have many different pages in our demo site for showing off our various components. There's no reason for all of them to be loaded at once though, if the user only needs to check out the inputs page. I've changed all of our pages to lazy modules only loaded when visited. This helps improve page load speeds and perceived responsiveness to the user. The app routing module now looks like this:

```
const routes: Routes = [
  { path: '', loadChildren: () => import('./landing-page/landing-page.module').then(m => m.LandingPageModule) },
  { path: 'alert', loadChildren: () => import('./dialogs/alert-page/alert-page.module').then(m => m.AlertPageModule) },
  {
    path: 'animations',
    loadChildren: () => import('./animations-page/animations-page.module').then(m => m.AnimationsPageModule)
  },
  {
    path: 'buttons',
    loadChildren: () => import('./forms/buttons-page/buttons-page.module').then(m => m.ButtonsPageModule)
  },
  {
    path: 'calendar',
    loadChildren: () => import('./forms/calendar-page/calendar-page.module').then(m => m.CalendarPageModule)
  },
  ...
];
```

## 8. Final results after Optimizations:

After doing these optimizations, this is what the bundle now looks like:

Screen Shot 2020-10-05 at 11 08 36 AM
Screen Shot 2020-10-05 at 8 38 46 AM

The lazy modules have been split into their own chunks, Moment has been reigned in, and Faker is no longer blowing up the size. The main module size is now less than half the size of the original. All of these optimizations together will help make a better user experience to anyone using our demo site, as well as reinforce best practices for developing efficient applications.


projects/swimlane/ngx-ui/tsconfig.lib.json
```diff
@@ -24,5 +24,5 @@
     "preserveWhitespaces": true,
     "enableIvy": false
   },
-  "exclude": ["src/test.ts", "**/*.spec.ts"]
+  "exclude": ["src/test.ts", "**/*.spec.ts", "src/lib/assets/icons/json/icons.json"]
```
marjan-georgiev, line 27: Do we still need this?

marjan-georgiev (approved):

projects/swimlane/ngx-ui/tsconfig.lib.json
luisb90: ahh good catch, we dont. removed!

PR #522, opened 2020-12-02
Stepper customization

joeveiga:
## Summary

- Feature(Stepper): Add `[large]` input to increase the default icon size.
- Feature(Stepper): Add `[trackBar]` input to make track bar optional.
- Feature(Stepper): Add `[progress]` input to show a progress indicator on the active step.
- Feature(Stepper): Add `[removeHighlight]` input to remove highlight color for the completed steps.
- Feature(Stepper): Add `[icon]` input to show an icon instead of the step number.
- Feature(Stepper): Add `.complete()` method to complete every step (including the last one).

image

## Checklist

- [ ] \*Added unit tests
- [x] \*Added a code reviewer
- [x] Updated the demo page

_\*required_

### DIFF
--- a/projects/swimlane/ngx-ui/CHANGELOG.md
+++ b/projects/swimlane/ngx-ui/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Feature(Stepper): Add `[large]` input to increase the default icon size.
 - Feature(Stepper): Add `[trackBar]` input to make track bar optional.
 - Feature(Stepper): Add `[progress]` input to show a progress indicator on the active step.
+- Feature(Stepper): Add `[removeHighlight]` input to remove highlight color for the completed steps.
 - Feature(Stepper): Add `[icon]` input to show an icon instead of the step number.
 - Feature(Stepper): Add `.complete()` method to complete every step (including the last one).
 
--- a/projects/swimlane/ngx-ui/CHANGELOG.md
+++ b/projects/swimlane/ngx-ui/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 # HEAD (unreleased)
 
+## 32.0.0 (2020-11-24)
+
+- Enhancement: ngx card component
+
 ## 31.1.0 (2020-11-10)
 
 - Fix: re-infer `json-editor` node types after model change
--- a/projects/swimlane/ngx-ui/CHANGELOG.md
+++ b/projects/swimlane/ngx-ui/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 # HEAD (unreleased)
 
+- Feature(Stepper): Add `[large]` input to increase the default icon size.
+- Feature(Stepper): Add `[trackBar]` input to make track bar optional.
+- Feature(Stepper): Add `[progress]` input to show a progress indicator on the active step.
+- Feature(Stepper): Add `[icon]` input to show an icon instead of the step number.
+- Feature(Stepper): Add `.complete()` method to complete every step (including the last one).
+
 ## 32.0.0 (2020-11-24)
 
 - Enhancement: ngx card component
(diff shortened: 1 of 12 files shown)


nartc (commented):
questions (non-blocking): LGTM but just quick question, why are there needs for using `cdr.markForCheck()` on every `@Input`?

joeveiga:
@nartc Thanks! Good question, I'm not sure (just copy/pasted one of them xD). My best guess would be that some of those properties can be updated by code (as opposed to template binding changes). See the [.first()](https://github.com/swimlane/ngx-ui/blob/02cb7c075369e53a22e7e2deb10091b5f616caf0/projects/swimlane/ngx-ui/src/lib/components/stepper/stepper.component.ts#L165) method in the `StepperComponent`, for example. Maybe would be better to move the `markForCheck()` call out of the properties an into those methods instead?

nartc:
suggestion (non-blocking): I see. It really depends on the intended usage of the StepperComponent. If the consumers query the StepperComponent by View/ContentChild then change StepperComponent's inputs from the consumers' component, then cdr.markForCheck might be needed. For this case, I'd try removing them. Since it looks like StepperComponent is being used with Input Bindings so StepperComppnent should be marked dirty properly without markForCheck unless there are things that I'm not aware of.

joeveiga:
@nartc That particular method (`first()`) is used like this in the demo page: https://github.com/swimlane/ngx-ui/blob/02cb7c075369e53a22e7e2deb10091b5f616caf0/src/app/components/stepper-page/stepper-page.component.html#L23
But yeah,, I'll try removing it and see. Thanks 🙏

nartc:
> @nartc That particular method (`first()`) is used like this in the demo page:
> 
> https://github.com/swimlane/ngx-ui/blob/02cb7c075369e53a22e7e2deb10091b5f616caf0/src/app/components/stepper-page/stepper-page.component.html#L23
> 
> 
> But yeah,, I'll try removing it and see. Thanks 🙏

I missed this. Yup, in this case, `cdr.markForCheck()` is needed.

Hypercubed:
I'm not sure where this pattern came from... I would discourage using `markForCheck` in inputs.  Most of the time it's at best redundant.

joeveiga:
@Hypercubed I agree. I removed some of them that seemed unnecessary. However, in the case of the `StepComponent`, I think we do need them for the current implementation to work correctly, unless I'm missing something. Some of the`StepComponent` properties can be set via template binding, as in:
```html
<ngx-stepper>
  <ngx-step [completeIcon]="myIcon"></ngx-step>
</ngx-stepper>
```
In which case this `this._cdr.markForCheck();` call doesn't do much:
https://github.com/swimlane/ngx-ui/blob/02cb7c075369e53a22e7e2deb10091b5f616caf0/projects/swimlane/ngx-ui/src/lib/components/stepper/step.component.ts#L35-L42
Or they can be set as part of the containing `StepperComponent`:
```html
<ngx-stepper [completeIcon]="myIcon">
  <ngx-step></ngx-step> <!-- inherits completeIcon property -->
</ngx-stepper>
```
And passed down to the steps using a `@ContentChildren(StepComponent)` reference (L112):
https://github.com/swimlane/ngx-ui/blob/02cb7c075369e53a22e7e2deb10091b5f616caf0/projects/swimlane/ngx-ui/src/lib/components/stepper/stepper.component.ts#L105-L120
In the second scenario, the change to the `completeIcon` property in the `StepComponent`, even though it's marked as an `Input()`, wouldn't be "change detected".. (or would it? :thinking: )

nartc:
> In the second scenario, the change to the `completeIcon` property in the `StepComponent`, even though it's marked as an `Input()`, wouldn't be "change detected".. (or would it? 🤔 )

That is correct. Angular Compiler depends on `[]` Input Binding syntax to actually check a component for Change Detection.
https://stackblitz.com/edit/ng-onpush-input?file=src%2Fapp%2Fhello-child.component.ts

nartc (approved):

duredhel13 (approved):

PR #547, opened 2021-01-25
Add extra @Output() for Date Picker to always emit value change

barbados-clemens:
## Summary

Add `inputChange` output for `ngx-date-time` to emit input changes even if the input value is not a valid date

## Reason
In certain cases, it's helpful to have an event that the value has changed, for example to run side effects in a [Component Store](https://ngrx.io/guide/component-store). even when a value isn't valid for for the desired Date/Datetime format.

## Note
This does not change the control value logic already in place, where it would only update the value if it valid or cleared. 

## Checklist

- [x] \*Added unit tests
- [x] \*Added a code reviewer
- [ ] Updated the demo page

_\*required_

### DIFF
--- a/projects/swimlane/ngx-ui/src/lib/components/date-time/date-time.component.spec.ts
+++ b/projects/swimlane/ngx-ui/src/lib/components/date-time/date-time.component.spec.ts
@@ -26,7 +26,7 @@ const LOCAL_HOUR = LOCAL_TIME.split(':')[0];
 const LOCAL_MIN = MOON_LANDING_DATE.toLocaleTimeString('en-US', { minute: 'numeric' });
 const LOCAL_AMPM = LOCAL_TIME.slice(-2);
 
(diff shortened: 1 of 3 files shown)


projects/swimlane/ngx-ui/src/lib/components/date-time/date-time.component.spec.ts
```diff
@@ -26,7 +26,7 @@ const LOCAL_HOUR = LOCAL_TIME.split(':')[0];
 const LOCAL_MIN = MOON_LANDING_DATE.toLocaleTimeString('en-US', { minute: 'numeric' });
 const LOCAL_AMPM = LOCAL_TIME.slice(-2);
 
-describe('DateTimeComponent', () => {
+fdescribe('DateTimeComponent', () => {
```
nartc, line 29: issue: remove focus

PR #563, opened 2021-03-04
Provide services in root injector

joeveiga:
## Summary

Provide singleton services in the root injector.

## Motivation

When singleton services that manage some sort of internal state are added to the module providers list vs the root injector, we can end up with multiple instances of the service (in lazy loaded modules, for example) resulting in an incorrect behavior.

## Example

The `DialogService` holds a collection of all the open dialogs (actually the collection is inherited from the abstract base class `InjectionRegistryService<T>`). The `destroyAll()` method iterates the collection and destroys all dialogs. Obviously having multiple instances of this particular service could be problematic:

```ts
// ...
dialogService__A.create({ /* ... */ });
dialogService__B.destroyAll(); // B's collection is empty, nothing to destroy here!
```

## Checklist

- [ ] \*Added unit tests
- [x] \*Added a code reviewer
- [ ] Updated the demo page

_\*required_

### DIFF
--- a/projects/swimlane/ngx-ui/CHANGELOG.md
+++ b/projects/swimlane/ngx-ui/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## HEAD (unreleased)
 
 - Bug(ngx-progress-spinner): Fix issue with later versions of SASS
-- Enhancement: Provide services in root injector (`DialogService`, `AlertService`, `NotificationService`)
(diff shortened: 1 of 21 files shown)


Hypercubed:
Changes look good to me... have you tested in out hybrid application?

joeveiga:
> Changes look good to me... have you tested in out hybrid application?

@Hypercubed Yeah I did some quick tests and it seems to be working fine. Dialogs, drawers, overlays, etc. behaving as expected.

Hypercubed (approved):
LGTM

PR #572, opened 2021-03-24
feat: add viewport intersection capabilities to the dropdown menu

duredhel13:
## Summary

The dropdown menu now determines if it's being partially hidden by the viewport and change direction upwards or downwards trying to be inside the viewport.

## Checklist

- [ ] \*Added unit tests
- [ ] \*Added a code reviewer
- [ ] Updated the demo page
- [x

https://user-images.githubusercontent.com/24899986/112237853-42a7c500-8c22-11eb-8ca5-e54375678e71.mov

] Included screenshots of visual changes

_\*required_

### DIFF
--- a/projects/swimlane/ngx-ui/src/lib/components/dropdown/dropdown-menu.directive.ts
+++ b/projects/swimlane/ngx-ui/src/lib/components/dropdown/dropdown-menu.directive.ts
@@ -1,4 +1,5 @@
-import { ElementRef, Directive } from '@angular/core';
+import { ElementRef, Directive, Inject, PLATFORM_ID, EventEmitter } from '@angular/core';
+import { InViewportDirective, InViewportService } from 'ng-in-viewport';
 
 @Directive({
   // tslint:disable-next-line:directive-selector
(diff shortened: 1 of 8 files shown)


projects/swimlane/ngx-ui/src/lib/components/dropdown/dropdown-menu.directive.ts
```diff
@@ -1,15 +1,25 @@
...
+import { InViewportDirective, InViewportService } from 'ng-in-viewport';
 
 @Directive({
   // tslint:disable-next-line:directive-selector
   exportAs: 'ngxDropdownMenu',
   selector: 'ngx-dropdown-menu',
   host: { class: 'ngx-dropdown-menu' }
 })
-export class DropdownMenuDirective {
+export class DropdownMenuDirective extends InViewportDirective {
```
Hypercubed, line 10: Is there a way to do this without inheritance?  Maybe using the service directly?

Hypercubed (commented):
One comment in the code.  As for imeplementation.. is is possible to use this service without the flash?  The dropdown first appears outside the window then pops into the window.  Maybe set opacity to 0 until the service has done it's thing?

projects/swimlane/ngx-ui/src/lib/components/dropdown/dropdown-menu.directive.ts
duredhel13: The best way I can figure think of to achieve the same result is by creating a separate directive that follows these patterns. They also involve inheritance but not in the same way as we are doing it now.

I'd have to try. But using the directive is a must, I'm afraid.

nartc: Is the Overlay from Angular CDK an option?

Hypercubed (approved):
Looks Good!

nartc (approved):

Hypercubed:
Just curious why we needed to change bahmutov/npm-install ?

duredhel13:
> Just curious why we needed to change bahmutov/npm-install ?

It was failing all the time, I manage to fix it. It seems that v1 is not being updated. The problem I was presenting was fixed in v1.4.5.

I used the latest version that was the solution.

Hypercubed (approved):

projects/swimlane/ngx-ui/src/lib/components/dropdown/dropdown-menu.directive.ts
k3nsei: Why not use it like below without inheritance.
https://github.com/k3nsei/ng-in-viewport/blob/develop/projects/example/src/app/page-lazy-images/page-lazy-images.component.html#L8
https://github.com/k3nsei/ng-in-viewport/blob/develop/projects/example/src/app/page-lazy-images/lazy-img.directive.ts#L61

PR #581, opened 2021-04-06
perf(button): reduce change detection cycles when the button is clicked

arturovt:
This PR reduces change detection cycles for the button component by replacing
`HostListener` with `Renderer.listen` outside of the zone. The actual click listener
doesn't change properties that could be reflected on the view, thus it doesn't need
Angular to run `ApplicationRef.tick()` each time the button is clicked. Note that the `HostListener`
wraps the actual listener under the hood into the internal Angular function which runs
`markDirty()` before running the actual listener (the decorated class method).

## Checklist

- [x] \*Added unit tests

### DIFF
--- a/projects/swimlane/ngx-ui/CHANGELOG.md
+++ b/projects/swimlane/ngx-ui/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Feature(CardComponent): add error outline
 - Fix(DropzoneComponent): addressing minor design review feedback
+- Enhancement(ButtonComponent): button clicks don't run change detection anymore
 
 =======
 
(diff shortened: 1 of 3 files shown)


nartc (approved):
LGTM

nartc:
@arturovt Thanks for the PR. Can you add some description of the change to https://github.com/swimlane/ngx-ui/blob/master/projects/swimlane/ngx-ui/CHANGELOG.md (the HEAD portion). Thanks!

arturovt:
@nartc thanks for the review! I've updated changelog. Sorry, I ain't familiar with your changelog format, ping me if it has to be changed.

PR #582, opened 2021-04-06
perf(dialog): reduce change detection cycles

arturovt:
This PR reduces change detection cycles for the dialog component by replacing
`HostListener` with `Renderer.listen` outside of the zone. The change detection should
be run only in specific cases, e.g. if the `esc` button has been clicked or the document
has been clicked and the target contains `dialog` class, but it shouldn't be run on each
button press and click event.
Note that the `HostListener` wraps the actual listener under the hood into the internal Angular
function which runs `markDirty()` before running the actual listener (the decorated class method).

## Checklist

- [x] \*Added unit tests

### DIFF
--- a/projects/swimlane/ngx-ui/src/lib/components/dialog/dialog.component.ts
+++ b/projects/swimlane/ngx-ui/src/lib/components/dialog/dialog.component.ts
@@ -3,7 +3,6 @@ import {
   Component,
   ElementRef,
   EventEmitter,
-  HostListener,
   Input,
   OnDestroy,
   OnInit,
@@ -11,8 +10,10 @@ import {
   ViewEncapsulation,
   Renderer2,
   TemplateRef,
-  ChangeDetectionStrategy
+  ChangeDetectionStrategy,
+  NgZone
 } from '@angular/core';
+import { ESCAPE } from '@angular/cdk/keycodes';
 import { coerceBooleanProperty, coerceNumberProperty } from '@angular/cdk/coercion';
 
 @Component({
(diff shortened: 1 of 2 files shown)


projects/swimlane/ngx-ui/src/lib/components/dialog/dialog.component.ts
```diff
@@ -151,15 +161,30 @@ export class DialogComponent implements OnInit, OnDestroy {
...
-    }
+  private setupEventListeners(): void {
+    this.ngZone.runOutsideAngular(() => {
+      const keydownEscListener = this.renderer.listen(this.element.nativeElement, 'keydown', (event: KeyboardEvent) => {
+        // We could've actually used `render.listener(element, 'keydown.esc')` and that would definitely work.
+        // Unfortunately, Angular delegates the event handling to the `KeyEventsPlugin` when it meets the `.` (dot)
+        // in the event name. The `KeyEventsPlugin` calls `ngZone.runGuarded` which triggers change detection anyway
+        // (no matter if we added event listeners in the root zone).
+        // In that case the change detection will be run only when the user clicks `esc` button.
+        if (event.keyCode === 27 && this.closeOnEscape) {
```
nartc, line 172: nitpick: let's use `Keycodes.ESCAPE` from AngularCDK. When they deprecate Keycodes and switch over to Keys, we will be able to make the change here as well.

nartc (commented):
LGTM. Just one small nitpick.

projects/swimlane/ngx-ui/src/lib/components/dialog/dialog.component.ts
arturovt: Thank you for bringing new knowledge! Updated the code.

arturovt:
I will also update the changelog as you asked in another PR. I'll wait until another PR will get merged since I will have to resolve conflicts.

PR #587, opened 2021-04-14
perf(drawer): reduce change detection cycles

arturovt:
This PR reduces change detection cycles for the drawer component by replacing
`HostListener` with `Renderer.listen` outside of the zone. The change detection should
be run only if the `Escape` button is clicked.
Note that the `HostListener` wraps the actual listener under the hood into the internal Angular
function which runs `markDirty()` before running the actual listener (the decorated class method).

The `DrawerService` has been updated to run the change detection only when the parent container is clicked.

## Checklist

- [x] Added unit tests

### DIFF
--- a/projects/swimlane/ngx-ui/CHANGELOG.md
+++ b/projects/swimlane/ngx-ui/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Feature(CardComponent): add error outline
 - Fix(DropzoneComponent): addressing minor design review feedback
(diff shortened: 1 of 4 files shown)


nartc:
@arturovt Hi, thanks a lot for your effort on working on these performance improvements. However, can you somehow make one PR for all of these change detection changes? That way, it is easier for me to review and us to keep track of the PR. Sorry for the inconvenience.

PR #617, opened 2021-05-21
Add validation when change on minDate or maxDate occurs

mario-alva:
## Summary

Add validation when change on minDate or maxDate occurs.

In this particular example we are using the value of the first component to be the value for the `[minDate]` input on the second component. 

When the first component value is changed, then the error message on the second component should go away, because now it is a valid date. 

That behavior was missing, because we did not have any validation happening after any change on the `minDate` or `maxDate`, the error message was always displayed, even if it was a valid date after the `minDate` value was changed.
image

image

## Checklist

- [x] \*Added unit tests
- [x] \*Added a code reviewer
- [ ] Updated the demo page

_\*required_

### DIFF
--- a/projects/swimlane/ngx-ui/CHANGELOG.md
+++ b/projects/swimlane/ngx-ui/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## HEAD (unreleased)
 
-- Enhancement: Add validation when change on `minDate` or `maxDate` occurs.
+- Enhancement: Add validation when change on `minDate` or `maxDate` occurs to the `date-time` component.
 
 ## 35.6.3 (2021-05-03)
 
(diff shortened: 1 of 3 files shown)


projects/swimlane/ngx-ui/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## HEAD (unreleased)
 
+- Enhancement: Add validation when change on `minDate` or `maxDate` occurs.
```
nelsonGuti, line 5: Can we add some detail regarding what component we are updating?

nelsonGuti (approved):

projects/swimlane/ngx-ui/CHANGELOG.md
mario-alva: Yes, we can. Added!

marjan-georgiev (approved):

PR #718, opened 2021-12-27
Date inputs form control

Hypercubed:
## Summary

`ngx-date-time` Changes:

* added axe tests for date-times
* add pop-up display for timezones
* Clean up `value` and add as input/output
* removed error displays
* added Validator and `validate` method
* added ControlValueAccessor `setDisabledState`
* added keyboard keys to open/close dialog
* added defined formats (same as time display)
* `(change)` now only emits dates

Other changes:

* fixed typing issue in unit tests
* renamed DATE_DISPLAY_TYPES.USER to DATE_DISPLAY_TYPES.TIMEZONE
* Cleaned up date/time input demo to use separate dates

Questions:
* should `(change)` event only emit dates?  Before it did.. usually.
* I removed the `Invalid Date` and `Out or range` errors text and suggest using form validation, should we keep the text... maybe only if the form is not used with a form control?

## Checklist

- [ ] \*Added unit tests
- [x] \*Added a code reviewer
- [ ] Updated the demo page

_\*required_


projects/swimlane/ngx-ui/src/lib/components/date-time/date-time.component.html
```diff
@@ -1,3 +1,47 @@
+<div class="ngx-date-time__container"
+  ngx-tooltip
```
haeminn, line 2: Do we want to set the tooltip on the container instead of on the actual input value? This will result in showing the tooltip on hovering anywhere in this component.

haeminn: Screen Shot 2022-01-03 at 9 03 25 AM
We should hide the tooltip when there is no value

haeminn (changes requested):

haeminn (approved):

marjan-georgiev (approved):

PR #811, opened 2022-05-17
schematics for ngx-ui

surya-pabbineedi:
## Summary
ngx-ui-schematics helps users to quickly scaffold all the required dependencies and configuration into angular cli projects.

image

Recording 2022-06-29 at 17 16 20

## Checklist

- [ ] \*Added unit tests
- [x] \*Added a code reviewer
- [ ] Updated the demo page

_\*required_


surya-pabbineedi:
@Hypercubed I have tried adding the `allowSyntheticDefaultImports` to tsconfig and I still see ajv error. Any thoughts on this? What is the issue with moment? 

Fixed issue with @angular/cdk and corrected @swimlane/ngx-ui-testing

Hypercubed:
@surya-pabbineedi Just tested this again... a couple of minor snags:

1) The same "allowSyntheticDefaultImports" issue as mentioned before (if we can't do anything about this let's just added it to the instructions)
2) ngx-ui still requires `legacy-peer-deps`, this is easily fixed by adding a `.npmrc` to the root.  If this can't be done via scematics then lets also add this to the docs.
3) Missing `resize-observer-polyfill` and `ngx-moment`.  Can we add this?
4) After doing `ng add` I  need to reinstall `npm i`, is that normal?

Other than these it worked!  Just a few more tweaks.

surya-pabbineedi:
Update code:

schematics now add .npmrc file with legecy-peer-deps settings (this will be added only if the file doesn't exists)
added missing packages
running npm install after adding all the deps

schematics

Hypercubed (approved):

PR #976, opened 2023-04-04
Feature Enhancement: Added Support for Displaying Tooltips on the ngx-select-option component

MarkColeman1:
## Summary

Added support for displaying tooltips on the `ngx-select-option` component. Previously, you could only display tooltips on the parent `ngx-select` component. 

## Checklist

- [ ] \*Added unit tests
- [x] \*Added a code reviewer
- [x] Updated the demo page

_\*required_


projects/swimlane/ngx-ui/src/lib/components/select/select-option.directive.ts
```diff
@@ -13,6 +14,17 @@ export class SelectOptionDirective implements SelectDropdownOption {
   @Input() name = '';
   @Input() value: any;
 
+  @Input() tooltipAlignment: AlignmentTypes = AlignmentTypes.center;
```
marjan-georgiev, line 17: Are all of these inputs needed? The tooltip directive should be providing them.

MarkColeman1: It does provide them but it does not work on `ngx-select-option`. Only works on the parent `ngx-select` and we need a way to display tooltips with custom templates for dropdown options.

drazafsky (approved):

PR #1025, opened 2024-01-11
Upgrade to Angular 17

surya-pabbineedi:
## Summary

- Upgraded demo project to Angular 17
- ngx-ui now works with Angular 17
- Fixed issues caused with select and unlockable inputs
- Select causes infinite change detection when used with template directive 
- Corrected unit tests

## Checklist

- [ ] \*Added unit tests
- [ ] \*Added a code reviewer
- [x] Updated the demo page

_\*required_


marjan-georgiev (approved):

PhilFlash:
Same problem with node with version 16.x but in netlify.

Best solution is to change node version on netlify with netlify GUI  @marjan-georgiev 
See [https://pratapsharma.com.np/set-node-version-netlify](https://pratapsharma.com.np/set-node-version-netlify)

Other solution is to create a file named .node-version in root directory with
```
18.19.0
```
See [https://docs.netlify.com/configure-builds/manage-dependencies/](https://docs.netlify.com/configure-builds/manage-dependencies/)

package.json
```diff
@@ -32,51 +32,50 @@
     "predeploy": "npm run build:lib && npm run build:deploy",
     "deploy": "angular-cli-ghpages --dir ./dist/ngx-ui --name=Swimlane --email=developer@swimlane.com",
     "prepublish:lib": "npm run package",
-    "publish:lib": "npm publish ./dist/swimlane/ngx-ui",
+    "publish:lib": "npm publish ./dist/swimlane/ngx-ui --ignore-scripts",
```
surya-pabbineedi, line 35: Publishing Angular library with --ignore-scripts to address NG0200 error

This npm publish command is configured with the --ignore-scripts flag to prevent the execution of post-publish scripts. The reason for this is to work around the NG0200 error (Circular dependency in DI) that occurs when compilationMode is set to 'partial' in Angular library compilation. Due to restrictions on publishing with 'full' compilationMode, --ignore-scripts is used to bypass these restrictions and successfully publish the library.

marjan-georgiev:
> Why to use full in compilationMode in projects/swimlane/ngx-ui/tsconfig.lib.prod.json
> 
> ```
> "angularCompilerOptions": {
>     "compilationMode": "full"
>   }
> ```
> 
> partial is recommended (and this is the default value when you create a library) See https://angular.io/guide/creating-libraries#publishing-libraries

We're running into compilation issues when building if we use partial compilation mode. We will use full until these issues are resolved. This means that going forward, the library will be only compatible with Ivy.

PR #1111, opened 2025-07-08
Date range calendar component

anjalitiwari285:
## Summary

Replace me with a description of changes. Include screenshots where appropriate.

## Checklist

- [ ] \*Added unit tests
- [ ] \*Added a code reviewer
- [ ] Added changes to `/projects/swimlane/ngx-ui/CHANGELOG.md` under HEAD (Unreleased)
- [ ] Updated the demo page
- [ ] Included screenshots of visual changes

_\*required_

PR #1116, opened 2025-07-16
Date range preset addition and bugfix

anjalitiwari285:
## Summary

Added more items in preset list to include week and quarter selection. Bug fix for invalid custom input string

## Checklist

- [x] \*Added unit tests
- [x] \*Added a code reviewer
- [ ] Updated the demo page
image

_\*required_

### DIFF
--- a/projects/swimlane/ngx-ui/CHANGELOG.md
+++ b/projects/swimlane/ngx-ui/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## HEAD (unreleased)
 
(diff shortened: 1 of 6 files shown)


jogaj:
@anjalitiwari285 I saw you removed a css class that was applying a max-width value to the component. Are u sure you had to remove it? Below is an example with an selected option that has a large label
image

You removed the class I am mentioning here https://github.com/swimlane/ngx-ui/pull/1111
image
The problem is that when you have more than one ngx-filter next to each other, and when the value selected is large, you can get a bad ux, for ex:

image

jogaj (approved):

PR #1139, opened 2025-09-01
Fix select options replaced when they were bound using the input options

jogaj:
## Summary

This is a fix to do not replace the value of the `options` input with the options added in the template. The issue was introduced in the following PR https://github.com/swimlane/ngx-ui/pull/1019

The new behavior will allow to overwrite the value of the `options` input only if the select has options defined in the template.

## Checklist

- [X] \*Added unit tests
- [X] \*Added a code reviewer
- [ ] Updated the demo page

_\*required_

### DIFF
--- a/projects/swimlane/ngx-ui/CHANGELOG.md
+++ b/projects/swimlane/ngx-ui/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## HEAD (unreleased)
 
-- Fix (`ngx-select`): Allow overwriting the value of the `options` input only if the select has its options defined by the template.
+- Fix (`ngx-select`): Only overwrite the `options` input if a value is provided in the template.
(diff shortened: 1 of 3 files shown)


projects/swimlane/ngx-ui/CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ## HEAD (unreleased)
 
-- Fix (`ngx-date-range-picker`): Fix end expression for presets values
+- Fix (`ngx-select`): Allow overwriting the value of the `options` input only if the select has its options defined by the template.
```
MarkColeman1, line 5: Can we update the verbiage to something a bit more concise like: "Only overwrite the `options` input if a value is provided in the template."

MarkColeman1 (changes requested):

PR #1197, opened 2026-04-15
(feat): Placeholder for additional actions

P-Anurag:
## Summary

- Added a placeholder beside the Close/Cancel button in the dialog header.
- We can now pass a custom template that will be rendered in the header. This is an optional property.
- Moved the template of the header-action to a separate html file for cleanliness.

https://github.com/user-attachments/assets/e5c4d234-8569-42a7-ad33-3e2bfe31d4cf

## Checklist

- [x] \*Added unit tests
- [x] \*Added a code reviewer
- [x] Updated the demo page

_\*required_

### DIFF
--- a/projects/swimlane/ngx-ui/src/lib/components/dialog/large-format/large-format-dialog-content.component.html
+++ b/projects/swimlane/ngx-ui/src/lib/components/dialog/large-format/large-format-dialog-content.component.html
@@ -58,6 +58,11 @@
     [imageTemplate]="imageTemplate"
     [dialogSubtitleTemplate]="dialogSubtitleTemplate"
   ></ngx-large-format-dialog-header-title>
+  <ng-container *ngIf="!stepperTemplate && !tabsTemplate">
+    <div *ngIf="largeFormatDialogHeaderContent" class="ngx-large-format-dialog-header-content">
+      <ng-container *ngTemplateOutlet="largeFormatDialogHeaderContent"></ng-container>
+    </div>
+  </ng-container>
   <ngx-large-format-dialog-header-action
     [dirty]="dirty"
     [actionTitle]="dialogActionTitle"
--- a/projects/swimlane/ngx-ui/src/lib/components/dialog/large-format/large-format-dialog-content.component.scss
+++ b/projects/swimlane/ngx-ui/src/lib/components/dialog/large-format/large-format-dialog-content.component.scss
@@ -63,6 +63,16 @@
       align-items: center;
       justify-content: space-between;
       padding: 0 2rem;
+
+      .ngx-large-format-dialog-header-content {
+        flex: 1;
+        margin: 0 var(--spacing-12);
+      }
+
+      ngx-large-format-dialog-header-title:has(+ .ngx-large-format-dialog-header-content),
+      .ngx-large-format-dialog-header-content + .ngx-large-format-dialog-header-action {
+        flex: 0 0 auto;
+      }
     }
 
     &__body {
--- a/projects/swimlane/ngx-ui/src/lib/components/dialog/large-format/large-format-dialog-content.component.spec.ts
+++ b/projects/swimlane/ngx-ui/src/lib/components/dialog/large-format/large-format-dialog-content.component.spec.ts
@@ -1,0 +1,204 @@
+import { ChangeDetectionStrategy, Component } from '@angular/core';
+import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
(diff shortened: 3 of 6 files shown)


steveblue (changes requested):
This behavior should act exactly like when the user adds tabs or the stepper to the header, although the feature here is custom content.

Create a content area that stretches from the header title/subtitle to close button. This allows users to define their own template. If they need "actions" on the right, they can style the template a few different ways to achieve that. 

Please follow naming conventions used elsewhere: largeFormatDialogHeaderContent would suffice for the name of the template.

Provide some examples in the page of left aligned, center aligned, right aligned content in the content area.

The documentation needs to outline that usage of this template will not work in tandem with the stepper or tabs.

projects/swimlane/ngx-ui/src/lib/components/dialog/large-format/components/large-format-dialog-header-action/large-format-dialog-header-action.component.ts
```diff
@@ -10,16 +10,7 @@ import {
 
 @Component({
   selector: 'ngx-large-format-dialog-header-action',
-  template: `
```
steveblue, line 13: Please avoid unnecessary changes. This makes upgrading even more difficult.

projects/swimlane/ngx-ui/src/lib/components/dialog/large-format/large-format-dialog-content.component.scss
```diff
@@ -63,6 +63,12 @@
       align-items: center;
       justify-content: space-between;
       padding: 0 2rem;
+
+      .ngx-large-format-dialog-header-content {
+        display: flex;
+        height: var(--header-height, 60px);
+        flex: 1;
```
steveblue, line 68: This seems like a potential problem for users that want to inject content that is not part of a flex layout. Maybe this should only have enough styling to ensure it fills the space properly.

projects/swimlane/ngx-ui/src/lib/components/dialog/large-format/large-format-dialog-content.component.scss
```diff
@@ -63,6 +63,12 @@
...
+      }
```
steveblue, line 75: It would be nice to at least provide some standard left and right spacing so the content area fits better in between the title and close button and not right up against.

projects/swimlane/ngx-ui/src/lib/components/dialog/large-format/large-format-dialog-content.component.html
```diff
@@ -58,6 +58,11 @@
     [imageTemplate]="imageTemplate"
     [dialogSubtitleTemplate]="dialogSubtitleTemplate"
   ></ngx-large-format-dialog-header-title>
+  <ng-container *ngIf="!stepperTemplate && !tabsTemplate">
+  <div class="ngx-large-format-dialog-header-content">
```
steveblue: There wasn't a way to inject this template like the stepper or tabs?

projects/swimlane/ngx-ui/src/lib/components/dialog/large-format/components/large-format-dialog-header-action/large-format-dialog-header-action.component.scss
```diff
@@ -1,7 +1,6 @@
 @use 'colors/colors' as colors;
 
 .ngx-large-format-dialog-header-action {
-  flex: 0 0 20%;
```
steveblue, line 4: We should try to avoid css changes like this, otherwise may have to wait for major version bump.

projects/swimlane/ngx-ui/src/lib/components/dialog/large-format/large-format-dialog-content.component.scss
P-Anurag: Removed display flex and height

projects/swimlane/ngx-ui/src/lib/components/dialog/large-format/large-format-dialog-content.component.scss
P-Anurag: Added margin of var(--spacing-12) on right and left

projects/swimlane/ngx-ui/src/lib/components/dialog/large-format/large-format-dialog-content.component.html
P-Anurag: As discussed offline, keeping the implementation as it is

steveblue (approved):

PR #1245, opened 2026-09-05
feat(ngx-select): inline tagging chip UX (commit, paste, keyboard, validator)

swim-with-rohit:
https://github.com/user-attachments/assets/d225eeb1-c27a-4b4b-a556-2c027944389f
## Summary

Adds **inline tagging** for `ngx-select` when there is no dropdown of options (`tagging && (disableDropdown || !options?.length)`).

Tagging-with-options, multi-select, and single-select behave as before.

### Behavior

- Commit tags with **Enter**, **Tab**, or **comma**
- Paste splits on comma / semicolon / tab / newline (also `\n` / `\t` escape text)
- Strips accidental HTML/entities from paste
- Commits unfinished text on blur or chip click
- Arrow keys select chips; Backspace/Delete remove (first Backspace highlights)
- Double-click a chip to edit it
- Long labels show a tooltip; optional `[taggingValidator]` for reject + invalid chip styling
- Fill appearance is the primary demo (`appearance="fill"`)

### API

- `taggingValidator?: SelectTaggingValidator` — **inline tagging only**
- `maxSelections` — unchanged shared input; also enforced on inline commit

## Checklist

- [x] Added unit tests
- [x] Added a code reviewer
- [x] Updated the demo page

## Test plan

- [ ] Demo → Tagging → **Basic**: still `<input>`, Enter adds tags, dropdown works
- [ ] Demo → Tagging → **Inline Tagging** (fill): Enter/Tab/comma, multi-value paste, blur commit, chip nav/edit, tooltip
- [ ] Demo → Autosize → **Inline Tagging** fill: textarea + chips align
- [ ] `[taggingValidator]` rejects bad tags (hint + no chip); existing bad values show invalid styling
- [ ] `yarn test:unit`
- [ ] `yarn test:prettier && yarn test:lint`

### DIFF
--- a/cypress/e2e/forms/selects.cy.ts
+++ b/cypress/e2e/forms/selects.cy.ts
@@ -202,18 +202,8 @@ describe('Selects', () => {
       cy.get('@CUT').find('.ngx-select-clear').first().click();
     });
 
-    it('supports free tagging entry and chip editing', () => {
-      cy.get('@SUT').find('ngx-select').eq(1).as('freeTagging');
-
-      cy.get('@freeTagging').find('textarea').type('one,two{enter}');
-      cy.get('@freeTagging').find('.ngx-select-input-option').should('contain.text', 'one');
-      cy.get('@freeTagging').find('.ngx-select-input-option').should('contain.text', 'two');
-
-      cy.get('@freeTagging').find('.ngx-select-input-option').contains('alpha').dblclick();
-      cy.get('@freeTagging').find('textarea').should('have.value', 'alpha');
-    });
-
     it('is keyboard accessible', () => {
+      cy.get('@CUT').clear();
       cy.get('@SUT').find('h4').contains('Basic').realClick();
 
       cy.get('@CUT').within(() => {
@@ -254,6 +244,17 @@ describe('Selects', () => {
 
       // Clears with backspace
       cy.focused().type('{backspace}{backspace}{backspace}'); // For some reason needs three backspaces in testing
+    });
+
+    it('supports free tagging entry and chip editing', () => {
+      cy.get('@SUT').find('ngx-select').eq(1).as('freeTagging');
+
+      cy.get('@freeTagging').find('textarea').type('one,two{enter}');
+      cy.get('@freeTagging').find('.ngx-select-input-option').should('contain.text', 'one');
+      cy.get('@freeTagging').find('.ngx-select-input-option').should('contain.text', 'two');
+
+      cy.get('@freeTagging').find('.ngx-select-input-option').contains('alpha').dblclick();
+      cy.get('@freeTagging').find('textarea').should('have.value', 'alpha');
     });
   });
 
--- a/cypress/e2e/forms/selects.cy.ts
+++ b/cypress/e2e/forms/selects.cy.ts
@@ -184,6 +184,7 @@ describe('Selects', () => {
       const text = 'Other';
 
       // TODO(ngx-ui-testing): support ngxFill for tagging
+      // Basic tagging (with options) still uses <input>; free tagging uses <textarea>.
       cy.get('@CUT').find('input').click().type(text).type('{enter}');
       cy.get('@CUT').ngxGetValue().should('equal', text);
 
@@ -199,6 +200,17 @@ describe('Selects', () => {
 
       cy.get('@CUT').find('.ngx-select-clear').first().click();
       cy.get('@CUT').find('.ngx-select-clear').first().click();
+    });
+
+    it('supports free tagging entry and chip editing', () => {
+      cy.get('@SUT').find('ngx-select').eq(1).as('freeTagging');
+
+      cy.get('@freeTagging').find('textarea').type('one,two{enter}');
+      cy.get('@freeTagging').find('.ngx-select-input-option').should('contain.text', 'one');
+      cy.get('@freeTagging').find('.ngx-select-input-option').should('contain.text', 'two');
+
+      cy.get('@freeTagging').find('.ngx-select-input-option').contains('alpha').dblclick();
+      cy.get('@freeTagging').find('textarea').should('have.value', 'alpha');
     });
 
     it('is keyboard accessible', () => {
--- a/cypress/e2e/forms/selects.cy.ts
+++ b/cypress/e2e/forms/selects.cy.ts
@@ -184,7 +184,6 @@ describe('Selects', () => {
       const text = 'Other';
 
       // TODO(ngx-ui-testing): support ngxFill for tagging
-      // Basic tagging (with options) still uses <input>; free tagging uses <textarea>.
       cy.get('@CUT').find('input').click().type(text).type('{enter}');
       cy.get('@CUT').ngxGetValue().should('equal', text);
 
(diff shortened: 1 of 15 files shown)


projects/swimlane/ngx-ui/src/lib/components/select/select-input.component.ts
```diff
@@ -15,6 +16,33 @@
...
+import { SelectTaggingValidator } from './select-tagging.interface';
+
+const TAG_SEPARATOR_PATTERN = /[,;\n\r\t]+/;
+/** Approx. truncated chip width (~18rem); shorter labels need no tooltip. */
+const CHIP_TOOLTIP_MIN_LENGTH = 32;
+
+function cleanTag(value: string): string {
+  return (
+    value
+      .replace(/<[^>]*>/g, '')
```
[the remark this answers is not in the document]
swim-with-rohit: False positive for this change.

`cleanTag` is only paste cleanup (strip accidental HTML / entities from Excel/Word), not an XSS sanitizer. The result is stored as a plain string chip value. Chip display still goes through Angular `[innerHTML]`, which DomSanitizer already handles — same pattern we already use for select options.

No security change needed here.

projects/swimlane/ngx-ui/src/lib/components/select/select-input.component.ts
```diff
@@ -15,6 +16,33 @@
...
+      .replace(/&nbsp;/gi, ' ')
+      .replace(/&amp;/gi, '&')
```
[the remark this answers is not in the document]
swim-with-rohit: False positive.

Unescaping `&amp;`, `&lt;`, etc. is intentional so pasted text becomes a normal readable tag. This isn’t a security boundary — just normalizing paste before we create chips.

projects/swimlane/ngx-ui/src/lib/components/select/select-input.component.ts
```diff
@@ -15,6 +16,33 @@
...
+      // eslint-disable-next-line no-control-regex -- strip control chars from pasted text
+      .replace(/[\u0000-\u001f\u007f]/g, '')
+      .replace(/[\u200b-\u200d\ufeff]/g, '')
+      .replace(/\u00a0/g, ' ')
+      .trim()
+  );
+}
+
+function plainChipLabel(option: SelectDropdownOption): string {
+  return `${option?.name ?? option?.value ?? ''}`.replace(/<[^>]*>/g, '').trim();
```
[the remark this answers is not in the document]
swim-with-rohit: False positive.

`plainChipLabel` only builds the tooltip title string. That text isn’t inserted as HTML, so the incomplete-tag / injection concern doesn’t apply here.

projects/swimlane/ngx-ui/src/lib/components/select/select-input.component.html
```diff
@@ -16,69 +16,97 @@
...
-              </span>
+              @if (isFreeTagging) {
+                @let tip = chipTooltip(option);
+                <span
+                  class="ngx-select-input-name"
+                  ngx-tooltip
+                  [tooltipTitle]="tip"
+                  [tooltipDisabled]="!tip"
+                  [tooltipType]="'tooltip'"
+                  [tooltipPlacement]="'top'"
```
surya-pabbineedi: Use plain attributes: 
```
tooltipType="tooltip" 
tooltipPlacement="top"
```

projects/swimlane/ngx-ui/src/lib/components/select/select-input.component.html
```diff
@@ -16,69 +16,97 @@
...
-                [ngTemplateOutletContext]="{ option: option }"
-                >
+              <ng-template [ngTemplateOutlet]="option.inputTemplate" [ngTemplateOutletContext]="{ option: option }">
               </ng-template>
             }
             @if (!option.inputTemplate) {
-              <span class="ngx-select-input-name" [innerHTML]="option.name || option.value">
-              </span>
+              @if (isFreeTagging) {
+                @let tip = chipTooltip(option);
```
surya-pabbineedi: chipTooltip(option) is also a method call inside `@for`.  runs plainChipLabel(regex) for every chip on every cycle. - compute tooltip in calcSelectedOptions.

projects/swimlane/ngx-ui/src/lib/components/select/select-tagging.interface.ts
```diff
@@ -0,0 +1,2 @@
+/** Optional per-tag validator. Return an error message to reject the tag, or null to accept. */
+export type SelectTaggingValidator = (value: any, selected: readonly any[]) => string | null;
```
surya-pabbineedi: Avoid using any

surya-pabbineedi (changes requested):

steveblue:
Shouldn't the typed text align vertically with the tag? In the video demo, the typed text appears to jerk down into the tab, whereas if they were vertically aligned, it may resolve more elegantly.

cypress/e2e/forms/selects.cy.ts
```diff
@@ -243,6 +245,17 @@ describe('Selects', () => {
       // Clears with backspace
       cy.focused().type('{backspace}{backspace}{backspace}'); // For some reason needs three backspaces in testing
     });
+
+    it('supports free tagging entry and chip editing', () => {
+      cy.get('@SUT').find('ngx-select').eq(1).as('freeTagging');
+
+      cy.get('@freeTagging').find('textarea').type('one,two{enter}');
```
steveblue: Please rename "Free Tagging" to "Inline Tagging" or "Inline Editable Tags", to be more explicit.

projects/swimlane/ngx-ui/src/lib/components/select/select-input.component.spec.ts
```diff
@@ -82,27 +83,28 @@ describe('SelectInputComponent', () => {
...
     describe('enter', () => {
       beforeEach(() => {
+        component.tagging = true;
         event.key = event.code = KeyboardKeys.ENTER;
       });
 
       it('should select value when not selected', () => {
         const spy = vi.spyOn(component.selection, 'emit');
         event.target.value = 'test';
-        component.onInputKeyUp(event);
```
steveblue, line 91: This seems potentially hazardous to change for existing modes.

projects/swimlane/ngx-ui/src/lib/components/select/select-input.component.ts
```diff
@@ -39,6 +54,8 @@ export class SelectInputComponent implements AfterViewInit, OnChanges {
   @Input() requiredIndicator: string | boolean;
   @Input() tabindex = 0;
   @Input() withHint = true;
+  @Input() maxSelections?: number;
```
steveblue, line 55: If the new feature incurs two inputs that are not used by another mode, could we combine or at least make more explicit if these inputs are only used by inline tags and none other?

projects/swimlane/ngx-ui/src/lib/components/select/select-input.component.ts
```diff
@@ -250,54 +358,238 @@ export class SelectInputComponent implements AfterViewInit, OnChanges {
...
+    // Free tagging then moves focus into the textarea via onFocus/onClick.
+    this.inputContainer?.nativeElement.focus();
+    if (this.isFreeTagging) this.focusInput();
   }
 
-  onClearTaggingInput(ev?: PointerEvent): void {
-    ev?.stopPropagation();
-    if (this.inputElement && this.inputElement.nativeElement) {
-      this.inputElement.nativeElement.value = '';
+  private onFreeTaggingKeyDown(event: KeyboardEvent): void {
```
steveblue, line 373: May need a check for  if (event.repeat) return; ?

steveblue:
Fill mode is the default we should using now. The demo in the video is based on the legacy form look and feel. Can you please make sure fill mode works seamlessly?

projects/swimlane/ngx-ui/src/lib/components/select/select-tagging.util.ts
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * This is paste/cleanup normalization — not an HTML security boundary.
+ * Safety for free tags comes from rendering them as text, not as HTML.
+ *
+ * Entity decoding is a single pass (e.g. `&amp;amp;` → `&amp;`).
+ * Must run on the full batch *before* splitting on `;`, otherwise `&amp;T` breaks apart.
+ */
+export function decodeFreeTagMarkup(value: string): string {
+  return value
+    .replace(/<[^>]*>/g, '')
```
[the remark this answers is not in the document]
swim-with-rohit: Addressed in the latest push: `decodeFreeTagMarkup` now (1) decodes entities in a **single** replace callback, (2) strips well-formed tags iteratively, then (3) removes any remaining `<`/`>` so incomplete markup like `<script` cannot survive (`<script` → `script`).

This is still paste/cleanup normalization — tags are rendered as plain text (`{{ chip.labelText }}`), not HTML. No CodeQL suppressions.

projects/swimlane/ngx-ui/src/lib/components/select/select-tagging.util.ts
```diff
@@ -0,0 +1,59 @@
...
+    .replace(/&nbsp;/gi, ' ')
+    .replace(/&amp;/gi, '&')
```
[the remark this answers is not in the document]
swim-with-rohit: Addressed in the latest push: entity decoding is now a **single-pass** regex with a replace callback (`decodeEntitiesOnce`), so `&amp;amp;` stays `&amp;` and there is no chained `&amp;` → `&` re-scan. Covered by unit tests.

steveblue:
Looking better. 

There is an odd UX disconnect though. When the user clicks on a tag the expectation should be the tag is editable, but the focus goes to the end of the field instead. The tag should be editable inline. This odd behavior with the tag hot swapped to the end is not standard UX and is not working as intended IMHO. I'm not sure the blue outline is helping this experience either, it takes away from the visibility of typing in the token.

The cancel buttons for each tag should be vertical align center.


Screenshot 2026-09-10 at 3 43 01 PM

swim-with-rohit:
> Looking better.
> 
> There is an odd UX disconnect though. When the user clicks on a tag the expectation should be the tag is editable, but the focus goes to the end of the field instead. The tag should be editable inline. This odd behavior with the tag hot swapped to the end is not standard UX and is not working as intended IMHO. I'm not sure the blue outline is helping this experience either, it takes away from the visibility of typing in the token.
> 
> The cancel buttons for each tag should be vertical align center.
> 
> Screenshot 2026-09-10 at 3 43 01 PM

Thanks, addressed in the latest push:

Double-click edits the chip in place (no more hot-swap to the end input)
Per-chip cancel (X) buttons are vertically centered

